PR #8, opened 2022-11-17
Created French translation

rene-coty:
I created the fr.po file + updated the LINGUAS file


GeopJr:
Thanks for creating this PR!

I'm going to merge it but for any future readers: Tooth is very much still in heavy development with almost no strings being final. I'll make sure to switch to weblate when they are ready.

PR #21, opened 2022-11-29
Add reply count

behDev:
I added this feature to show the count of replays. If one is replayed, 1 is shown and if more than 1, +1 is shown.

Screenshot from 2022-11-29 20-31-36

### DIFF
--- a/data/ui/widgets/status.ui
+++ b/data/ui/widgets/status.ui
@@ -257,16 +257,16 @@
           </object>
         </child>
         <child>
-              <object class="ToothWidgetsAvatar" id="avatar">
-                <property name="size">48</property>
-                <property name="valign">start</property>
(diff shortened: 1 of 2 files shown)


GeopJr:
I added 2 changes:

- Use horizontal margin if there are replies

before:
screenshot of the reply button with the "1" label. It horizontally overflows over the button style

after:
screenshot of the reply button with the "1" label. It no longer horizontally overflows over the button style

- Handle the case of 0 replies (useful for live updates / when a "delete post" button gets added and replies_count goes < 1)

If they look good to you, I'll proceed with merging this PR!

behDev:
Awesome! merge it

GeopJr:
Merging, thanks!

PR #22, opened 2022-12-03
fix: Copy a string to the clipboard

behDev:


GeopJr:
I added 3 commits:

- Removed the 'Share' section from the menu bar (as the circular button replaces it (+ the "Copy Profile Link" item never worked))
- Added a null check to Gdk.Display.get_default (I doubt we will ever encounter it but since it's nullable it doesn't hurt to check it)
- Made it copy the full handle and not the relative one (aka "@username@instance" always instead of just "@username" if you are on the same instance as that account)

If everything looks good, I'll go ahead and merge this!

behDev:
Everything is good, please merge it.

GeopJr:
Thanks, merging!

PR #23, opened 2022-12-03
build: Remove post_install.py

fortysixandtwo:
meson.build uses gnome.post_install()
so the post installation script is not needed.


GeopJr:
LGTM, thanks!

fortysixandtwo:
Thanks for having picked up tootle maintenance!

PR #24, opened 2022-12-03
desktop: Indicate we can handle StartupNotify

fortysixandtwo:
As GTK apps do this just fine let's have a splash
under phosh.


GeopJr:
LGTM, thanks!

PR #25, opened 2022-12-05
poll support

jcamposz:
This code allows to poll in Tooth. You can view, and vote but not create a poll. 
Viewing:
imatge
Voting:
imatge

### DIFF
--- a/data/gresource.xml
+++ b/data/gresource.xml
@@ -61,6 +61,7 @@
         <file>ui/widgets/list_item.ui</file>
         <file>ui/widgets/list_editor_item.ui</file>
         <file>ui/widgets/compose_attachment.ui</file>
+        <file>ui/widgets/votebox.ui</file>
         <file>ui/dialogs/new_account.ui</file>
         <file>ui/dialogs/compose.ui</file>
         <file>ui/dialogs/main.ui</file>
--- a/data/ui/widgets/status.ui
+++ b/data/ui/widgets/status.ui
@@ -219,7 +219,7 @@
                           </object>
                         </child>
                         <child>
-                              <object class="ToothWidgetsVoteBox" id="poll_spoiled">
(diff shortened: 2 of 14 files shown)


GeopJr:
Thanks for adding polls support!

I did some changes:
- Let the spoiler button handle the poll hidden status
- Changed the design to match both Mastodon's and hig's:

1. Replaced the box + buttons with listbox and actionrows
2. Used a gradient as the progressbar for the poll results
3. Placed the vote button, "people voted" and "expires in" in the same place as mastodon
4. Added the people "voted label"
5. Changed the "expired" labels humanizers to better match the context (aka "expired 5h ago" "expired yesterday") (+ handle weeks and days)
6. Made the vote button insensitive if there's no choice done

Screenshot of the Tooth app of a post that has poll results. User info and poll info are censored. At the top there's a check icon with the label "Post Results". At the end of the row below it, there's a globe icon and the label "Dec 4". There's also a small warning icon some rows below. There are 3 poll options. The first one is at 64.00% (title). The background is accent colored (blueish) at .5 alpha. The action bar is 64% filled with that background color. Similarly for the other two options at 16.00% and 20.00% but instead of an accent colored background there's a slightly lighter version. Below it there are the labels "25 voted" and "expired 11h ago"
Screenshot of the Tooth app of a post that has poll. User info is censored. At the top there's a message bubble icon with the label "*blank* Mentioned you". At the end of the row below it, there's an envelope icon and the label "10m". There are 2 poll options. Each option has a radio button (unselected) on the left. The first one has the title "test2t" and the second one the label "t". Below it theres an insensitive vote button and the labels "0 voted" and "23h left".

If everything looks ok to you, I'll go ahead and merge this!

jcamposz:
Nice modifications.

GeopJr:
Merging, thanks!

PR #26, opened 2022-12-06
meson: depend on 0.56 instead of 0.59

kop316:
Hello!

Your README says Tooth needs meson 0.50, but the build requires 0.59. To make this functional for 0.56, I added a check to see what meson it is on, and changed how the post-install scripts are run.


meson.build
```diff
@@ -153,8 +153,16 @@ executable(
 subdir('data')
 subdir('po')
 
-gnome.post_install(
+# gnome.post_install() is available since meson 0.59.0
+# Distributions use their own tooling (e.g. postinst, triggers, etc)
+# so it is okay if the post_install() is not run on distro builds
+m_ver = meson.version().split('.')
+if m_ver[0].to_int() > 0 or m_ver[1].to_int() > 58
```
eli-schwartz, line 160: No!!!

Please, use `meson.version().version_compare('>=0.59.0')`. :)

eli-schwartz (changes requested):

meson.build
kop316: fair enough, fixed!

eli-schwartz: Thanks. :)

kop316: Thank you! I didn't know about that before. I agree, it is a nicer function.

GeopJr:
Thanks for creating this PR!

I'm honestly slightly unsure about whether I should move forward with this.

- As part of modernizing Tootle's (/Tooth's) scripts I adopted parts of the default Vala template from GNOME Builder [which moved to 0.57.0 *specifically* to replace the postinstall script with the gnome module](https://gitlab.gnome.org/GNOME/gnome-builder/-/commit/f086e66d525cd93cb9a4514bc0f9858b9c0c0314) (and later to 0.59.0). (+ the script got removed on https://github.com/GeopJr/Tooth/pull/23)
- Instead of relaxing the validation, I'd rather fix it (or at least keep it as a reminder (in the case of screenshots for example) until Tooth gets closer to release).

The version listed on the readme if definitely wrong however.

Let me know about your thoughts on this!

kop316:
> Thanks for creating this PR!
> 
> I'm honestly slightly unsure about whether I should move forward with this.

No worries! What I will say is this: you are ultimately upstream, so the decision is yours. I am **not** offended at all if you decide that what I have shouldn't be included.

> 
>     * As part of modernizing Tootle's (/Tooth's) scripts I adopted parts of the default Vala template from GNOME Builder [which moved to 0.57.0 _specifically_ to replace the postinstall script with the gnome module](https://gitlab.gnome.org/GNOME/gnome-builder/-/commit/f086e66d525cd93cb9a4514bc0f9858b9c0c0314) (and later to 0.59.0). (+ the script got removed on [build: Remove post_install.py #23](https://github.com/GeopJr/Tooth/pull/23))

I do this on my GUI apps so that I can maintain compatibility with PureOS. I daily drive a Librem 5 with PureOS, so this allows me to build on it and test my apps. Byzantium is based on Debian Bullseye and have meson `0.56` installed,, so this is how I make sure I can build on PureOS.

I personally feel it is a good balance between backwards compatibility for PureOS while making sure newer distros can take advantage of the newer features. However, I understand that maintaining backwards compatibility can be a a pain.

> 
>     * Instead of relaxing the validation, I'd rather fix it (or at least keep it as a reminder (in the case of screenshots for example) until Tooth gets closer to release).

I agree that fixing it should be the ultimate goal. I did this because when I build a Debian package, it errors out if any test fails (I have a half baked Deb packaging here: https://github.com/kop316/Tooth/tree/wip/Deb ). You may also get an issue from folks saying tests fail if it isn't relaxed.

Maybe this is a stop gap until the validation is fixed?

On a sidenote: Are you interested in the Debian Packaging? It isn't finished, but if you are, I can polish it up and make a separate PR.

> 
> 
> The version listed on the readme if definitely wrong however.

Agreed.

> 
> Let me know about your thoughts on this!

kop316:
Also FWIW: I did submit an issue in PureOS to update Meson to 0.59: https://source.puri.sm/Librem5/OS-issues/-/issues/301

GeopJr:
I see, thanks for the info!

Since this is outside of our control I guess I can merge this. (I will probably revert this however when Tooth becomes release-ready and published on Flathub (not anytime soon)).

Similarly, I'll merge the tests change (but I'll also revert it as soon as screenshots are ready (the ui is not finished yet)).

> Are you interested in the Debian Packaging?

Thanks for working on it! I'd rather not maintain it however so I'll have to deny.

Merging!

kop316:
> I see, thanks for the info!
> 
> Since this is outside of our control I guess I can merge this. (I will probably revert this however when Tooth becomes release-ready and published on Flathub (not anytime soon)).

Thank you for doing that!

> 
> Similarly, I'll merge the tests change (but I'll also revert it as soon as screenshots are ready (the ui is not finished yet)).

That sounds reasonable. i agree, once the issues are done, that change can be reverted.

> 
> > Are you interested in the Debian Packaging?
> 
> Thanks for working on it! I'd rather not maintain it however so I'll have to deny.

No worries!

> 
> Merging!

fortysixandtwo:
> I do this on my GUI apps so that I can maintain compatibility with PureOS. I daily drive a Librem 5 with PureOS, so this allows me to build on it and test my apps. Byzantium is based on Debian Bullseye and have meson `0.56` installed,, so this is how I make sure I can build on PureOS.

maybe a bit late to the party, but the post install thing is not needed on distro builds (well I can only really speak for Debian based - but iirc an arch maintainer also confirmed these schema compilation etc is also triggered when upgrading packages using pacman).
Also `DESTDIR` will often get set during package builds making the post_install.py (see the [script](https://github.com/GeopJr/Tooth/blob/main/build-aux/meson_post_install.py#L12)) -- and also the gnome.post_install() (see the [documentation](https://mesonbuild.com/Gnome-module.html#gnomepost_install))` -- effectively a NOOP.

This is why I proposed #23 a few days ago ;)

Anyway I don't think this hurts, but wanted to leave some comments because my changes were reverted :cry:

eli-schwartz:
> maybe a bit late to the party, but the post install thing is not needed on distro builds (well I can only really speak for Debian based

I think the only distro -- or Unix, non-linux OS package manager -- that doesn't use DESTDIR is NixOS, which uses --prefix for that instead (don't ask).


> Also `DESTDIR` will often get set during package builds making the post_install.py (see the [script](https://github.com/GeopJr/Tooth/blob/main/build-aux/meson_post_install.py#L12)) -- and also the gnome.post_install() (see the [documentation](https://mesonbuild.com/Gnome-module.html#gnomepost_install))` -- effectively a NOOP.

However, the use of DESTDIR doesn't make the `gnome.post_install()` function a NOOP. It still evaluates at build time, thus raising the minimum meson version. It just doesn't run at install time in the end. This PR is about the minimum meson version.

PR #29, opened 2022-12-12
Background Mode

behDev:
Currently background mode works by hiding the window. Also in this mode we need to send notification when a new toot is sent

### DIFF
--- a/data/ui/dialogs/preferences.ui
+++ b/data/ui/dialogs/preferences.ui
@@ -67,12 +67,12 @@
         <child>
           <object class="AdwPreferencesGroup">
             <property name="title" translatable="yes">Behavior</property>
-            <property name="sensitive">0</property>
             <child>
               <object class="AdwActionRow">
                 <property name="title" translatable="yes">Autostart</property>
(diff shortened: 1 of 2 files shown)


GeopJr:
LGTM, thanks!

Should `work-in-background` be true by default? I assume the flatpak builds will ask the user if they want to allow it to run in the background anyway.

Also, should there be a sidebar item for quitting the app? If someone wants to actually close the app they'd have to either turn off the setting and close the window or kill the app.

behDev:
> Should work-in-background be true by default? I assume the flatpak builds will ask the user if they want to allow it to run in the background anyway.

We can ask this from user during login. I don't know anything about flatpak

> Also, should there be a sidebar item for quitting the app? If someone wants to actually close the app they'd have to either turn off the setting and close the window or kill the app.
edit: perhaps time to introduce a menu popover? At least for keyboard shortcuts and quitting?

i think adding a checkbox on the popover for quitting is a good idea and access to this is faster, or adding an option in preferences to ask about quit when closing the window.

GeopJr:
Thanks for the feedback!

Merging!

PR #34, opened 2022-12-20
Fix POTFILES and regenerate po files

rffontenelle:


GeopJr:
Looks good to me! Thanks!

PR #35, opened 2022-12-20
Add Brazilian Portuguese translation

rffontenelle:
This is based on #34, so please check that first.

### DIFF
--- a/po/LINGUAS
+++ b/po/LINGUAS
@@ -1,4 +1,5 @@
 de_DE
 es
 fr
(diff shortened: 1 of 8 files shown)


GeopJr:
Thanks for creating this PR!

The same comment from #8 applies here as well:

> I'm going to merge it but for any future readers: Tooth is very much still in heavy development with almost no strings being final. I'll make sure to switch to weblate when they are ready.

PR #36, opened 2022-12-21
feat: lists re-write

GeopJr:
[Screen recording of the Tootle GTK app in dark mode on the lists page. The user first creates a new list with the name "test4". The list gets instantly added next to the others. Then the user clicks the edit button which make a settings window visible, toggles between the radio buttons and then closes the settings window. Lastly the user deletes the last two lists and see them instantly disappear.](https://user-images.githubusercontent.com/18014039/209019960-66cf552e-1e18-4d68-b9f9-36b129578461.webm)

Done:

- Add lists
- Delete lists
- View lists
- First page of list settings
- Second page of list settings (member management)
- Save settings
- Add user to list


GeopJr:
I think it's pretty straight-forward now:

[Screen recording of the Tootle GTK app in dark mode on the lists page. The user is on a test account's profile. They first click the kebab menu at the headerbar to reveal the menu popup without the "Lists" entry. The user then follows the test account and goes back to the menu to see that the "Lists" entry is now there. A modal dialog shows up with a list of the available lists. There are two random named ones. The user first adds the test account to the second one, then the first and then removes it from the second. Afterwards, they hover over the minus button (remove button) to reveal its tooltip text and then close the dialog. The user then visits the "Lists" page, revealing the two lists. Then clicks the first one, waits for its timeline to load and then goes back. Then clicks the edit button, changes its name and replies policy and closes the dialog revealing that the name change saved since it changed the row title. Then clicks the edit button again, revealing that the other changes got saved and visits the members tab. There the user switches the test account mentioned before, to off, closes the window and clicks it again revealing that theres no longer a members tab since there are no members.](https://user-images.githubusercontent.com/18014039/209241536-108e3d91-8fb3-468a-b14e-bdc52b0c7ea3.webm)

PR #37, opened 2022-12-22
Add a Color Scheme Preference

### LINKED ISSUE #17: Bug Report - support dark theme in the app
#### What should happen:
__

tooth should support black theme

#### What happened instead:
__
tootle supported dark theme, but moving to tooth, there is no option to choose dark theme. Though choosing system wide dark theme switches tooth also to dark theme. It may be a good idea to expose this option in tooth also. Though not very strong about it as well.

#### How to reproduce:
1. __
2. __

#### Screenshots (if applicable):
__

#### Build context:
 - **Distro:** __ mobian
 - **App version:** __ git master
 - **Build type:** Flatpak / Built from source / Built by a third-party maintainer - built from source
https://people.debian.org/~praveen/tooth/README (my personal repo)

#### Logs:
*(Terminal logs are often invaluable. If you can, launch the app from terminal with `dev.geopjr.tooth` and paste the output here)*

#### Additional info (if needed):
__

Diego-Ivan:
Fixes #17 

In this merge request, I implement a Color Scheme switcher that allows changing from this system, light and dark. I will do a little self review too.

[Grabación de pantalla desde 2022-12-22 00-19-08.webm](https://user-images.githubusercontent.com/76420970/209070864-730dc79a-4baf-4a80-bcd6-7174bf09dd34.webm)

### DIFF
--- a/data/dev.geopjr.tooth.gschema.xml
+++ b/data/dev.geopjr.tooth.gschema.xml
@@ -1,14 +1,20 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <schemalist>
 
+  <enum id="dev.geopjr.tooth.ColorScheme">
+    <value value="0" nick="system"/>
+    <value value="1" nick="light"/>
+    <value value="2" nick="dark"/>
+  </enum>
+
 	<schema path="/dev/geopjr/tooth/" id="dev.geopjr.tooth" gettext-domain="dev.geopjr.tooth">
 
 		<key name="active-account" type="s">
 			<default>''</default>
 		</key>
-		<!-- <key name="dark-theme" type="b">
-			<default>false</default>
-		</key> -->
+		<key name="color-scheme" enum="dev.geopjr.tooth.ColorScheme">
+			<default>'system'</default>
+		</key>
 		<key name="autostart" type="b">
 			<default>false</default>
 		</key>
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -109,6 +109,10 @@ namespace Tooth {
 				error (msg);
 			}
 
+			var style_manager = Adw.StyleManager.get_default ();
+			ColorScheme color_scheme = (ColorScheme) settings.get_enum ("color-scheme");
+			style_manager.color_scheme = color_scheme.to_adwaita_scheme ();
+
 			set_accels_for_action ("app.about", ACCEL_ABOUT);
 			set_accels_for_action ("app.compose", ACCEL_NEW_POST);
 			set_accels_for_action ("app.back", ACCEL_BACK);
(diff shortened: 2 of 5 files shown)


data/dev.geopjr.tooth.gschema.xml
```diff
@@ -1,14 +1,20 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <schemalist>
 
+  <enum id="dev.geopjr.tooth.ColorScheme">
+    <value value="0" nick="system"/>
+    <value value="1" nick="light"/>
+    <value value="2" nick="dark"/>
+  </enum>
+
```
Diego-Ivan, line 9: Here I add an enumeration to the gschema, which takes certain values and nick to set it accordingly

src/Services/Settings.vala
```diff
@@ -30,5 +30,36 @@ public class Tooth.Settings : GLib.Settings {
 	void init (string key) {
 		bind (key, this, key, SettingsBindFlags.DEFAULT);
 	}
-
 }
+
+public enum Tooth.ColorScheme {
+	SYSTEM,
```
Diego-Ivan, line 36: This is an enumeration with the values that the application will handle. SYSTEM, LIGHT and DARK. It two methods:

* to_string (): Used to display it in the preferences window. They're translatable too!
* to_adwaita_scheme (): Which converts the values to their equivalent in Adw.ColorScheme

src/Dialogs/Preferences.vala
```diff
@@ -42,4 +49,58 @@ public class Tooth.Dialogs.Preferences : Adw.PreferencesWindow {
...
+	private void on_scheme_changed () {
+		var selected_item = (ColorSchemeListItem) scheme_combo_row.selected_item;
+		var style_manager = Adw.StyleManager.get_default ();
+
+		style_manager.color_scheme = selected_item.adwaita_scheme;
+		settings.color_scheme = selected_item.color_scheme;
+	}
+}
+
+public class Tooth.ColorSchemeListModel : Object, ListModel {
```
Diego-Ivan, line 62: Both ColorSchemeListModel and ColorSchemeListItem are requisites to use Adw.ComboRow, as we need an item and a list model. I preferred to use this implementation as I felt using Gtk.StringList and then converting the value to an enum and setting it would be a little 

The ColorSchemeListItem has two properties, name and adwaita_scheme, which are used to display the name of the enumeration to the Adw.ComboRow, and adwaita_scheme to set the value to the style_manager when it is obtained.

src/Application.vala
```diff
@@ -109,6 +109,10 @@ namespace Tooth {
 				error (msg);
 			}
 
+			var style_manager = Adw.StyleManager.get_default ();
```
Diego-Ivan, line 112: Here the value from the settings is obtained as a ToothColorScheme, and its adwaita equivalent is given to the style manager.

GeopJr:
LGTM, thanks for the PR and the explanation!

PR #38, opened 2022-12-22
Slightly faster window resizing

Diego-Ivan:
Small fix. Currently, the window when being resized looks a little laggy, this MR makes it look faster. 

To implement this, the settings service saves its changes until the application is closed, instead of saving them when any property is changed, which makes in general the tweaking of settings in the preferences window slightly faster too.


GeopJr:
LGTM, thanks!

PR #39, opened 2022-12-27
Build System Cleanup

Diego-Ivan:
This merge request implements:

* The Build.vala.in file has been replaced by a .vapi file and a C header file
* The meson.build file has been split into various files for the src/ subdirectory
* Add the Vala SDK Extension, which provides autocompletion for Flatpak GNOME Builder and Flatpak VS Code
* Add a G_LOG_DOMAIN variable


GeopJr:
cc: @fortysixandtwo @kop316

meson.build
```diff
@@ -9,25 +9,22 @@ project(
...
+config.set_quoted('PREFIX', get_option('prefix'))
+config.set_quoted('NAME', 'Tooth')
+config.set_quoted('WEBSITE', 'https://github.com/GeopJr/tooth')
+config.set_quoted('SUPPORT_WEBSITE', 'https://github.com/GeopJr/tooth/issues')
+configure_file(output: 'config.h', configuration: config)
+
+add_project_arguments (
+  '--include','config.h',
+  '-DGETTEXT_PACKAGE="@0@"'.format(meson.project_name()),
+  language: 'c'
```
fortysixandtwo, line 29: I know `c` was already used before, but I'm wondering whether it shouldn't be `vala` instead 
as it is mentioned as a supported language string in the [docs](https://mesonbuild.com/Reference-manual_functions.html#project)

src/Application.vala
```diff
@@ -219,11 +238,12 @@ namespace Tooth {
...
-				copyright = Build.COPYRIGHT,
-				debug_info = Build.SYSTEM_INFO,
-				developers = developers,
-				artists = artists,
-				translator_credits = Build.TRANSLATOR != " " ? Build.TRANSLATOR : ""
+				copyright = COPYRIGHT,
+				debug_info = print_info (),
+				developers = DEVELOPERS,
+				artists = ARTISTS,
+				// translators: Write you Name<email> here :)
```
fortysixandtwo, line 245: typo: s/you/your 
:)

fortysixandtwo (commented):
Only really found two nits, lgtm

meson.build
Diego-Ivan: Vala is supported, but those arguments are for the C compiler, as it asking to include the C header that was generated using `configure_file`. On the other hand, we will use that C Header in Vala through the VAPI that is included in the project files

GeopJr:
Sorry for the wait!

I brought this up-to-date and organized it slightly more (one meson.build per folder)

PR #43, opened 2023-01-03
Added Dutch translation

Vistaus:
🙂


GeopJr:
Thanks for creating this PR!

The same comment from #8 applies here as well:

> I'm going to merge it but for any future readers: Tooth is very much still in heavy development with almost no strings being final. I'll make sure to switch to weblate when they are ready.

PR #45, opened 2023-01-06
main: Use the over transition for the leaflet

Kekun:
The slide transition reveals the lack of separator between the pages and the headerbar border, which looks odd. The over transition fixes that. Also the slide transition doesn't denote a back gesture as well as the over transition.


GeopJr:
LGTM, thanks!

PR #46, opened 2023-01-07
Fix header text wrapping in sidebar

bleakgrey:
Sometimes the text in sidebar header is wrapped even if there is room for the text. This PR expands the text box and adds a slight vertical padding to make the layout a bit more neat.

| Before  | After |
| ------------- | ------------- |
| Screenshot from 2023-01-07 14-20-48 | Screenshot from 2023-01-07 14-20-31 |

### DIFF
--- a/data/ui/views/sidebar/view.ui
+++ b/data/ui/views/sidebar/view.ui
@@ -3,6 +3,7 @@
   <template class="ToothViewsSidebar" parent="GtkBox">
     <property name="width_request">300</property>
     <property name="orientation">vertical</property>
+    <property name="hexpand">false</property>
 
     <child>
       <object class="AdwHeaderBar">
@@ -62,8 +63,9 @@
                         <object class="GtkBox">
                           <property name="orientation">vertical</property>
                           <property name="valign">center</property>
+                          <property name="hexpand">true</property>
+                          <property name="spacing">6</property>
                           <child>
-                            <!-- FIXME: Wrapping -->
                             <!-- <object class="ToothWidgetsEmojiLabel" id="title"> -->
                             <object class="GtkLabel" id="title">
                               <property name="label">Account Name</property>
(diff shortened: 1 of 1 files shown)


GeopJr:
Thanks for creating this PR!

Looks good to me, I just reverted the FIXME comment removal as it's about the (commented out) emoji label that currently looks like this:

screenshot of the tooth gtk app zoomed in the sidebar header. The user name doesn't wrap and there's many emojis overflowing the sidebar on the x axis.

bleakgrey:
Ah I see. Packing emojis as separate widgets in a horizontal Box doesn't seem like a good idea - this approach doesn't work with word wrapping at all.

I think I had some research on drawing custom images directly on Gtk.Label widgets using `pangocairo`. I never managed to make a prototype, but maybe I could try again if I find it. Would you be interested in that?

GeopJr:
Sure! Other alternatives I've considered are:

- `Gtk.TextBuffer.insert_paintable` https://valadoc.org/gtk4/Gtk.TextBuffer.insert_paintable.html
- Using webviews (which will be needed sooner or later if activitypub servers add more formatting options) which would also add support for animated emojis. But it depends on libsoup3.

bleakgrey:
IMHO the former should be used in a pinch as it is not intended to be used in such quantities (GtkTextView makes more sense as a singular widget in the Composer dialog due to its rich editing capabilities).

As for the latter, webviews defeats the idea of using a native client. Besides, I can already tell that spawning a webview instance every time rich markup is expected (which is usually always and everywhere) will skyrocket the RAM usage, which is already abnormal. Might as well just use a browser 😄 

Animated emojis is a valid point though. I'll see what we can do about them.

GeopJr:
I see, thanks for the info!

For what it's worth, Fractal uses a custom label widget with children: https://gitlab.gnome.org/GNOME/fractal/-/blob/main/src/components/label_with_widgets.rs

bleakgrey:
I absolutely love the idea of using a separate widget for each embedded image. Fractal uses Pango's shape attribute - exactly what I had in mind. 

It seems to use these attributes to render custom glyphs in the Label's markup, get their pixel positions and use them for the embedded widgets. I think. Maybe. I wish I had more experience in Rust, it's mostly Greek to me :/

PR #50, opened 2023-01-09
Refactor Add Account dialog

bleakgrey:
This PR makes the following changes to the dialog:
- Flatten the header bar
- Replace vanilla `Gtk.Entry` widgets with their `Adw.EntryRow` variants
- Change wording on some screens
- Hide the auth token entry unless manual authorization is explicitly requested by the user
- Fix the Network service so that error messages are properly passed to the dialog

| Before  | After |
| ------------- | ------------- |
| Screenshot from 2023-01-09 22-24-04  | Screenshot from 2023-01-09 22-23-52  |
| | Screenshot from 2023-01-09 22-38-44 |
|  Screenshot from 2023-01-09 22-40-20  |  Screenshot from 2023-01-09 22-42-21  |
| Screenshot from 2023-01-09 22-40-29 | Screenshot from 2023-01-09 22-38-54 |

Feedback and testing appreciated 😄

### DIFF
--- a/data/style.css
+++ b/data/style.css
@@ -9,6 +9,10 @@
   font-weight: bold;
 }
 
+headerbar.flat .title {
+	opacity: 0;
+}
+
 .attachment, .attachment button, flowboxchild {
 	padding:0px;
 	margin:0px;
--- a/data/style.css
+++ b/data/style.css
@@ -9,7 +9,7 @@
   font-weight: bold;
 }
 
-headerbar.flat .title {
+headerbar.flat.no-title .title {
 	opacity: 0;
 }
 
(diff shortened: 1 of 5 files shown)


GeopJr:
Looks great, thanks!

I did some changes:
- There was a critical log when `guess_backend` was called (prior to access_token being assigned)
- Limited the flat headerbar title opacity style to `headerbar.flat.no-title` (so it doesn't get applied if we need a flat headerbar with visible title in the future)
- I bound `auth_page.description` and `use_auto_auth` because there was a small delay between the auth code entry appearing and the description changing


I think the instance entry should placeholder should change from `https://` to something more descriptive like "Server URL" (mastodon app) or "instance" (tusky)

bleakgrey:
LGTM, changed the placeholder text to "Server URL"

GeopJr:
Awesome, merging!

PR #52, opened 2023-01-11
(Blocked) Drop Widgets.Background

bleakgrey:
`Widgets.Background` was created because `Gtk.Picture` lacked the desired display mode for the Profile header image. In GTK 4.8, `Gtk.Picture` got new `content-fit` property which supersedes this widget.

This PR probably shouldn't be merged _right now_ because:
- it would make Tooth depend on GTK >= 4.8 which some distros are yet to adopt
- valac 0.56 seems to be completely ignorant of the `content-fit` property (it can't be set programmatically)

Still, it should be kept in mind that this widget is no longer needed and we may need to remove it soon.


GeopJr:
I'll turn it into a draft so we don't accidentally merge it!

(But also I'm not going to wait forever for distros to update to GTK >= 4.8 after a flatpak gets released)

Issue #53, opened 2023-01-11
Leaking Status widgets


bleakgrey:
### Describe the bug

All Status widgets refuse to get destroyed. As a side-effect, timeline-based pages like Profile or Conversations don't get destroyed too, and entity cache keeps the statuses and images they contained in memory forever.

I'm pretty sure it didn't happen in Tootle (can't tell for sure though, it didn't compile on my device), which means this is caused by the new functionality in the Status widget.

I couldn't pinpoint the exact cause yet - this issue needs testing.

### Steps To Reproduce

1. Open many timelines in the main window
2. Navigate back
3. Observe the terminal output - there are no messages like `Destroying Status widget`

### Instance Backend

Mastodon

### Operating System

Pop!_OS 22.04 LTS

### Package

Flatpak

### Troubleshooting information

flatpak: true
version: 2.1.0
gtk: 4.8.3 (4.8.3)
libadwaita: 1.2.0 (1.2.0)

### Additional Context

When destroyed properly, Status widgets emit a debug message specified in Status.vala (line 139). Additionally, unused API.Status entities get freed by entity cache with a message specified in AbstractCache.vala (line 39).

All of these aren't present in `main`.

bleakgrey:
Update: I just noticed that `Views.Lists` is leaking too - there is no proper destruction message in the terminal.

GeopJr:
can confirm they get destroyed correctly on Tootle:

```
** Message: 20:28:07.795: Status.vala:76: Destroying Status widget
** Message: 20:28:07.797: Thread.vala:18: Destroying Thread
** Message: 20:28:07.797: ContentBase.vala:32: Destroying ContentBase
** Message: 20:28:07.797: Base.vala:53: Destroying base Conversation
```

```
** Message: 20:28:23.278: ContentBase.vala:32: Destroying ContentBase
** Message: 20:28:23.278: Base.vala:53: Destroying base Lists
```

My current speculation is that I probably messed it up during the navigation rethink commits:

https://github.com/GeopJr/Tooth/commit/5ba32dfa8357426f671c7183455401d399e44f40
&
https://github.com/GeopJr/Tooth/commit/c73927b08d1cb95df8ca660b976a738179d498f1

bleakgrey:
Both commits seem harmless enough to me, it can't be navigation. I'd check if new property bindings or something like that was added to the Status widget itself.

GeopJr:
[...]

PR #54, opened 2023-01-12
Fix leaking Status widgets

bleakgrey:
This PR partially f!xes #53. It's based on the `BindingGroup` patch proposed there, but without breaking spoilers.


bleakgrey:
I'm not sure why, but valac fails to capture reference to the `spoiler_stack` properly. Using a notify instead seems to work.

GeopJr:
Awesome, much cleaner than the "property" solution, thanks!

PR #55, opened 2023-01-12
Add meson profiles

bleakgrey:
This PR adds a `profile` option to the meson configuration. If is set to `development` during compilation, the app will:
- Add `.devel` class to the main window
- Use alternative nightly icon
- Modify About dialog to display branch name and revision hash instead of a version (which makes sense in production builds only)

Additionally, a development Flatpak manifest is added with [Vala SDK](https://github.com/flathub/org.freedesktop.Sdk.Extension.vala) enabled for syntax highlighting in GNOME Builder.

Screenshot from 2023-01-12 11-56-42

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
       options: --privileged
     steps:
     - uses: actions/checkout@v3
-    - uses: flatpak/flatpak-github-actions/flatpak-builder@master
+    - uses: flatpak/flatpak-github-actions/flatpak-builder@v5
       with:
-        bundle: "dev.geopjr.tooth.flatpak"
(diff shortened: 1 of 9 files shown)


bleakgrey:
We can use this option to disable some debug information in production builds, such as widget destruction logs.

GeopJr:
Looks good to me!

I did the following changes:

- CI should build devel
- Added a check if git is available 
- There was a warning about the missing "check" flag on run_command, so I added it
- And made install.sh build devel

(and merged upstream)

Are there any other profiles that could be added in the future? If not, should devel be a bool instead?

bleakgrey:
I can't think of any other profile, so yeah, a bool makes sense.

GeopJr:
Merging!

PR #59, opened 2023-01-14
workflow: Build for aarch64

TheEvilSkeleton:
Related to https://github.com/GeopJr/Tooth/issues/41.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,9 +10,25 @@ jobs:
     container:
       image: bilelmoussaoui/flatpak-github-actions:gnome-43
       options: --privileged
+    strategy:
+      matrix:
+        arch: [x86_64, aarch64]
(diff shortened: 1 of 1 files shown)


TheEvilSkeleton:
I'll be building it in my fork to avoid getting approved. Once done, I'll let you know.

GeopJr:
Looks good to me!

I limited the qemu and docker actions to aarch64 (and bumped the qemu one to v2).
If you still have changes to do, I'll just turn this into a draft until it's ready!

TheEvilSkeleton:
I have no changes to make. But wow, this is unbearably slow :sweat_smile:. I do certainly believe that we should keep it so you can have testers. Of course, it's entirely slow.

GeopJr:
Nothing we can really do I guess, still better than nothing!

Merging, thanks!

PR #63, opened 2023-01-20
feat: instance info

GeopJr:
Instance info is important for the editor dialog (new toot)

I'll leave this up for a bit in case I missed any compat functions or nullable/instance-specific properties


GeopJr:
Blocked:

instance_info needs to be serialized for secret, however it already causes too many problems and a refactor is needed. Secret should hold only the absolutely necessary and static info about the account and not the whole Entity.

PR #70, opened 2023-02-09
Add `Ctrl+Q` quit shortcut

DaPigGuy:
[GNOME HIG](https://developer.gnome.org/hig/reference/keyboard.html)


GeopJr:
Looks good to me, thanks!

PR #130, opened 2023-03-27
fix(meson): add distro option to disable post_install

carlwgeorge:
As mentioned in an earlier comment, Linux distributions use their own packaging tooling to accomplish tasks such as compiling glib schemas and updating the icon cache.  This adds a "distro" meson option to allow distribution packagers to easily disable those tasks during the meson build.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -10,6 +10,7 @@ project(
 )
 
 devel = get_option('devel')
+distro = get_option('distro')
 
 # Setup configuration file, accessed via vapi/build.vapi
 config = configuration_data()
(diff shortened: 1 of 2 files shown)


GeopJr:
LGTM, thanks!

eli-schwartz:
IMHO this is not needed -- the design of the post_install() method was specifically to cater to distros here, because it does not run when distros run their own packaging commands -- i.e. $DESTDIR is defined.

carlwgeorge:
While creating an RPM package for version 0.1.0, I remember getting some kind of error about one of the post_install-related commands not being available, which led me to believe that that post_install() needed a way to be skipped.  I'm not able to reproduce it now.  The behavior seems to match what you're describing.  I'm not sure what I did incorrectly before that triggered the error.  This could probably be reverted.

PR #137, opened 2023-03-29
feat(Composer): Add initial autocomplete support

bleakgrey:
This PR enables account handle autocomplete for the Compose dialog. 

image

The implementation depends on `gtksourceview-5`, the library that makes interactive text suggestions possible in other GNOME apps like Text Editor or Builder. I felt reluctant to introduce new dependencies, but I figured this one was probably already available on the host system since we target GNOME.

In its current state, it's more of a proof-of-concept (although being a useful thing to have already). If viable, we could iterate on it further:

- [ ] Consider using different styling of the popup window
- [x] Autocomplete instance emojis via a separate provider
- [ ] Replace `Gtk.Label` with `Tuba.LabelWithWidgets` once it's ready
- [ ] Implement spell checking?
- [ ] Utilize `SourceView`'s syntax highlighting. Pleroma supports Markdown/HTML posts, so maybe we could support multiple content types too?

Tested against GNOME 44 SDK with `gtksourceview-5` 5.8.0.


GeopJr:
Looks great so far!

I adapted it to libsoup3 & added Tag.search for hashtags - not sure if adding hashtag support in the same provider is the best idea so I left it at that

Megaf:
> Lgtm.
> 
> Each provider is supposed to autocomplete just one type of data, so they are better off being separate classes. I'll add ~tag and~ instance emoji providers as separate PRs.
> 
> I got carried away while refactoring and ended up making a HashtagProvider as well. Here's how it looks like now: image

This is looking great! Would it be possible to change the colour of the FediverseID/Hastag once it is "selected/validaded"? So, we know we are actually sending a message to that FediverseID.

GeopJr:
Added emojiprovider (might be too strict, I made it check if it starts with but we can switch to contains)

I think it can make it on 0.2.0!

- Does it need a custom stylescheme for dark mode? On my machine at least, it's white on dark mode
- Does it use anything from 5.8 or will <= 5.6 also work? Since almost everyone ships 5.6 at most right now (Debian Sid, Fedora, Alpine, Snap core22, Freebsd (5.4))

[Screencast from 2023-03-31 21-16-02.webm](https://user-images.githubusercontent.com/18014039/229198973-1966c3ce-31ff-403f-9c04-51dd6d04fc37.webm)

bleakgrey:
I completely forgot that style schemes exist, but now it should respect the app color theme.

As for the library version... well, I _did_ manage to compile this PR with `5.6.0`... but I had to remove something that definitely shouldn't be a breaking change.

GeopJr:
ah https://gitlab.gnome.org/GNOME/gtksourceview/-/issues/308 fixed in 5.7.1, guess we will add it to troubleshooting when we move to that version :shrug:

GeopJr:
Made completion provider use its own image_cache with 1min maintenance since the images have already been freed by the time the user scrolls back in all providers:

before

[Screencast from 2023-04-02 11-52-33.webm](https://user-images.githubusercontent.com/18014039/229342910-ebc20625-9e6e-4ec4-98f3-26f1b270f568.webm)

after

[Screencast from 2023-04-02 11-51-39.webm](https://user-images.githubusercontent.com/18014039/229342925-9218bd69-2ccc-469c-ad6d-12c2029c2910.webm)

GeopJr:
Merging! I split some of the raised issues into their own ones: #153, #154

Thanks!

Issue #12, opened 2022-11-18
libsoup3


GeopJr:
Currently Tooth doesn't do a great job at handling errors - especially ones from libsoup that mostly depend on the instance. At the time of writing this, my account causes a segfault when it starts querying my instance.

With the excuse of switching to libsoup3, it should also handle more errors.

PR #138, opened 2023-03-29
feat: libsoup3

GeopJr:
TODO:

- [x] test everything that does any request
- [x] investigate random crash
- [x] re-do voting

What changed?

- Tuba.Request is now a wrapper for Soup.Message since it cannot longer inherit from it
- Session queue has now been replaced by `send_async`
- Cancelling is handled using cancellable on Tuba.Request
- `response_body` has been replaced with InputStreams and is no longer part of Soup.Message
- Request's Soup.Message is only finalized when exec is called - don't assume it exists prior to that
- network.queue is internal and should only be used directly with Soup.Messages + its InputStream needs to be manually handled (see ImageCache)

fix: #12


GeopJr:
Merging, hopefully I tested everything

PR #141, opened 2023-03-30
fix: 4.10.0 deprecations

GeopJr:


GeopJr:
The only one left is:

```
../src/Widgets/VoteBox.vala:78.17-78.37: warning: `Gtk.Widget.get_style_context' has been deprecated since 4.10
   78 |                 row.get_style_context ().add_provider (provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
```
which might not have a replacement yet: https://gitlab.gnome.org/GNOME/gtk/-/issues/5254 (while `gtk_style_context_add_provider_for_display` is mentioned as non-deprecated, `Gtk.StyleContext` as a whole is marked as deprecated right now)

GeopJr:
I think it's good to go if the CI finishes

Issue #112, opened 2023-03-23
The App-Icon (Could we talk about it?)


fabiscafe:
### Describe the request

cc @bertob because it's [your work](https://github.com/GeopJr/Tuba/pull/104) and I need your experience here. :smile: I'm pretty sure that there was a lot of talk about it in the background, that I don't know about :smile: 

To be honest. I think the icon is, while a good artwork, not a good-working app-icon. 

![grafik](https://user-images.githubusercontent.com/18744080/227269882-e283b904-8eb8-4c5b-a6df-d5be7c40fac7.png)

This is a view of my fav-bar. Tuba is very visible. Not only because the icon itself is huge, but it's also bright, almost shiny thanks to the gradients - the other icons are much more subtle, flat with only minor highlights in comparison. Tuba also is one of the few ones that are not made using basic shapes and in a more detailed fashion, that is hard to see on smaller resolutions. This leads to a very out of place looking style on the desktop. In a short form I'd say "it feels like it is to much for my eyes".

While I can not come up with a different idea I wanted to open this to point this out, hopefully start another(?) talk about it. Not as a bug, it's not a bug and I don't want to push this down in any way. As I said in the beginning. It's good artwork.

### Implementation Details

- [ ] This should be an option in settings.
- [ ] This should be only available to some fediverse backends. (Include which ones on the above field).
- [ ] This is client-only (and shouldn't sync with the instance).
- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

GeopJr:
While I cannot comment on everything, the shininess, yellow color and complexity are part of tuba's shape and materials:

<details><summary>(Image of a tuba)</summary>

![an image of a tuba](https://user-images.githubusercontent.com/18014039/227281997-f37ad39d-b3d9-4429-932a-06d04d09a2f4.png)

</details>

you are more than welcome to take a look at past iterations (including some with less complexity and blue color instead of yellow) over at the app icon design room: https://matrix.to/#/#appicondesign:gnome.org

PR #151, opened 2023-04-02
app icon: adjust size

bertob:
As pointed out in #112, the current icon is a bit too large. This makes it a little less tall, bringing it more in line with other icons in terms of visual weight, while still looking like a tuba.

image


GeopJr:
LGTM, thanks!

Issue #192, opened 2023-04-17
[Request]: Contextual action to be able to quickly add/remove someone to one or more lists


nekohayo:
### Describe the request

Lists can be a bit of a pain to manage when you have to go into the Lists UI, grab each individual list, click its edit button, click the Members tab, and then try to add/remove accounts there (and for some reason, it seems unable to show the current members of lists on Mastodon, as far as I can tell).

It would be very nice if you could right-click anyone's profile or statuses/posts/toots, and add/remove that person to one of your lists. Mastodon's web UIs allow that from their "…" meatballs menu when you are following someone. It would be a quicker alternate way to manage lists, particularly when one person would be relevant to multiple lists at once.

### Implementation Details

- [ ] This should be an option in settings.
- [ ] This should be only available to some fediverse backends. (Include which ones on the above field).
- [ ] This is client-only (and shouldn't sync with the instance).
- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

GeopJr:
That's already the case but it seems to have broken during the libsoup3 migration:

- Go to someone's profile
- Click the headerbar menu
- Select "Lists"
![image](https://user-images.githubusercontent.com/18014039/232562976-dc95bdee-358c-489c-86d2-2f6f0ba52d03.png)

- Add/remove them from your lists
![image](https://user-images.githubusercontent.com/18014039/232561957-7130a4f4-adff-45eb-b644-dfe444f68a8a.png)

(Obviously that's broken in 0.2.0)

I don't think adding that option to the post context menu is ideal as it's unrelated to it and we are going to reach mastodon levels of overwhelming context menus:
![image](https://user-images.githubusercontent.com/18014039/232562800-2992eda1-888c-4bf9-8245-5f7eca28dcf6.png)

GeopJr:
(re-opening to track it with the PR)

GeopJr:
Thanks for raising this issue it should be fixed as of ce5344d0cac0aea878cafdb6b21b0f439c33aa4d!

PR #193, opened 2023-04-17
fix(network): re-use functions and parsers

GeopJr:
fix: #192 

During the libsoup3 migration we switched to inputstreams. The main issue is that once read it cannot be reused. This PR adds a parser parameter to supply our own parser instead of creating a new one. Additionally since all the array network functions were using the same code from `get_array_mstd`, I just made them call it.

I'm leaning towards making them use a user-provided parser only and get rid of handling inputstreams directly.


GeopJr:
Ended up using parsers for everything

Issue #159, opened 2023-04-05
[Request]: Move @username inline with name


bertob:
### Describe the request

The @username is currently below the name, which makes posts taller than they need to be, creates unnecessary whitespace on the right, and doesn't look great visually. I'd move it up next to the name, like most other clients do.

![image](https://user-images.githubusercontent.com/1908896/230046804-a4c09b7d-7458-410f-ab8a-3fcab58bff7a.png)

Elk:

![image](https://user-images.githubusercontent.com/1908896/230045158-2086ef6f-5fe7-4f75-9c5c-2e4c3b32497e.png)

Phanpy:

![image](https://user-images.githubusercontent.com/1908896/230046470-10e9b921-f5a2-427b-a87f-d0fd17eb8c7b.png)

Ivory:

![image](https://user-images.githubusercontent.com/1908896/230047510-dd9fed99-3e99-45c0-8a0c-062ba55d79d0.png)

### Implementation Details

- [ ] This should be an option in settings.
- [ ] This should be only available to some fediverse backends. (Include which ones on the above field).
- [X] This is client-only (and shouldn't sync with the instance).
- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

GeopJr:
Since this will change the layout of the status, it might be better to combine it with #62 as to avoid future revisits:

additional post actions (like Edit or Delete) are in the context menu, it might be better to have them as a button since they are relatively hidden.

- Elk places the button next to the date, but also doesn't display post indicators (visibility, pinned, edited icons)
- Mastodon places it next to the post actions, but doesn't display a dedicated bookmark button. - Mastodon mobile app places it at the top right but doesn't display indicators and the date is next to the handle

bertob:
I think both the top right and bottom right in the action bar would work layout-wise, semantically I think I have a slight preference for top right.

GeopJr:
I went with top right!

The main issue with the inline handle is wrapping and spacing. For example the following post:

![Screenshot from 2023-04-10 12-45-37](https://user-images.githubusercontent.com/18014039/230885538-f47fe386-47ba-44e0-9a52-4139a5234c65.png)

If the content moves higher, it doesn't leave much space for the button. If the name + handle + indicators + button also move higher, the button will reach the container limits

![Screenshot from 2023-04-10 13-05-59](https://user-images.githubusercontent.com/18014039/230886105-c52b6cc1-61b0-4cff-a76d-8f601eba253f.png)

[...]

PR #217, opened 2023-04-25
icons: scale inline icons smaller

jimmac:
See https://github.com/GeopJr/Tuba/issues/159#issuecomment-1513230618 for context


GeopJr:
LGTM, thanks!

I'm going to tag @bertob for spacing input:

Non expanded:

top is 3, bottom is 6 (current value)

Screenshot from 2023-04-25 19-08-29

Expanded:

current spacing

Screenshot from 2023-04-25 19-08-58

decreased spacing

image

even more decreased spacing

image

GeopJr:
Sounds good to me!

Turns out I missed another indicator, [`tuba-pin-symbolic`](https://github.com/GeopJr/Tuba/blob/main/data/icons/scalable/actions/tuba-pin-symbolic.svg), could you redraw it too?

PR #218, opened 2023-04-25
Add Swedish Translation

bittin:


bittin:
However the Settings panel is not translated when i tested locally and compiled it with meson but everything else is, but maybe you can merge it directly from Weblate or know why this happens?

2023-04-25-120943_1920x1080_scrot

bittin:
@GeopJr sure i can close it and wait for the proper merge from Weblate, just added it so i could test locally then i remembered i could do my own fork to test things out

Issue #126, opened 2023-03-27
[Request]: do not scroll on new posts


stacyharper:
### Describe the request

Dunno if this should be considered a feature or a bug. While streaming post is on, new posts make my view to scroll on new posts.

Ideally I think it shouldn't scroll while prepending post. So that I could go to the top, go AFK some time, come back and just scroll to the top until I hit the head.

### Implementation Details

- [ ] This should be an option in settings.
- [ ] This should be only available to some fediverse backends. (Include which ones on the above field).
- [X] This is client-only (and shouldn't sync with the instance).
- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

Issue #228, opened 2023-05-04
More specialized desktop and mobile layouts


bertob:
The navigation model is currently not ideal on either desktop or mobile:
- On desktop it's weird to have sidebar items for some things, and view switcher items on the right for others
- On mobile it's weird to have a huge sidebar for what would otherwise be a normal primary menu

Using the new AdwBreakpoints it should be possible to have independent layouts for the two cases, which would help improve the layout for both. Here's a strawman proposal:

- For desktop I'd move everything to the sidebar, so you'd essentially have three new things in there (below Home): Notifications, Conversations, and Search
- For mobile I'd make the sidebar a normal primary menu on the right

GeopJr:
Looks great so far!

Here's a quick attempt at the desktop one (no breakpoints yet) (https://github.com/GeopJr/Tuba/tree/experiment/feat/redesign)

![Screenshot from 2023-05-05 17-02-58](https://user-images.githubusercontent.com/18014039/236480100-225f841d-5ef0-4532-8482-55b5bdeec7c1.png)
![Screenshot from 2023-05-05 17-02-54](https://user-images.githubusercontent.com/18014039/236480108-c011a8d3-0a40-457b-b7b0-ceea31022587.png)

The sidebar needs a complete re-write so I didn't bother for now

(It's probably a bit early to ask but) should the top left sidebar button open the account switcher or the current users' profile and instead put the switcher in the menu next to it?

should the scroll to top button switch to the left side? (it's removed in that branch)
and should the compose / new post button be visible on all views or just home?

imhemish:
Just to say, KDE's Tokodon also somewhat looks like this
![image](https://user-images.githubusercontent.com/74466492/236516559-b849689a-6fc8-42c6-8913-15f9219e8722.png)
Looks awesome!

bertob:
Nice, that's a great start! A few notes:

- The background colors on both panes should be darker, not sure where you'd get the correct values though, cc @exalm 
- The sidebar is too wide, so it looks empty. I'd make it about 220px wide.

I pushed my initial mockup here, including a mobile layout: https://gitlab.gnome.org/Teams/Design/other-app-mockups/-/blob/master/tuba/tuba.png

> should the top left sidebar button open the account switcher or the current users' profile and instead put the switcher in the menu next to it?

I was thinking it'd open the account switcher, but good point about needing to have the link to your own profile somewhere. Maybe it's fine if that's available only from the account switcher?

[...]

PR #234, opened 2023-05-08
feat(Timeline): entity queue

GeopJr:
fix: #126

### How does this work?
- If the user's scroll position is near the top: new posts get prepended
- If the user's scroll position is not near the top: new posts get added into an array and when the user scrolls to near the top, they get prepended

### DIFF
--- a/src/Views/ContentBase.vala
+++ b/src/Views/ContentBase.vala
@@ -5,6 +5,7 @@ public class Tuba.Views.ContentBase : Views.Base {
 	public GLib.ListStore model;
 	protected ListBox content;
 	private bool bottom_reached_locked = false;
+	protected signal void reached_close_to_top ();
 
 	public bool empty {
 		get { return model.get_n_items () <= 0; }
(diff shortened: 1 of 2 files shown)


darkdragon-001:
UX wise [Tusky on Android](https://github.com/tuskyapp/Tusky) is pretty good. One feature I like there in combination with this one here is that you can scroll to the top via click on the current menu item. Also it will automatically start from top after some time / after restarting the app?

GeopJr:
> One feature I like there in combination with this one here is that you can scroll to the top via click on the current menu item

This has been suggested before and the main point was that such pattern is not found in any other GNOME app and might distort its (the view switcher's) expectations in other apps. Like GNOME Software or Music. There's already a scroll to top button anyway. This needs more thought but probably after #228 

> Also it will automatically start from top after some time / after restarting the app?

(I assume on Tuba, not tusky) This PR shouldn't cause any noticeable changes to users (apart from the scrolled window not moving around when new posts get added) so everything will work like it did before: On start, it starts at the top and at no point does it automatically change the scrolled position without the user triggering it (either through the scroll to top button or manually scrolling to the top)

Just to go a bit deeper on how this PR works in case it wasn't that clear:

(zoom in if the text is not visible)
prior to this PR:
Untitled-2023-05-11-0420

notice how the post you are reading "moved" below when a new post got added

after this PR:
Untitled-2023-05-11-0420(1)

PR #239, opened 2023-05-10
feat(Profile): Add joined date

be4zad:


GeopJr:
LGTM, thanks!

I made some changes:

- use DateTime parsing instead of doing it manually
- Format the date as `MONTH D, YYYY`
- Mark strings as translatable
- Add an icon as prefix (so it kinda looks different from the other fields, to avoid confusion if a user creates a "Joined" field themselves)
- Moved it below the user-submitted fields

Screenshot from 2023-05-10 20-55-39

Let me know if you have any comments on any of them, otherwise I'll merge when the CI is done!

Issue #244, opened 2023-05-11
[Request]: Peertube support


RasheedAZ:
### Describe the request

When a Mastodon post has a video attached to it, Tuba can play it if the user clicks on the video. However, for video uploads from Peertube accounts, only a link to the video page is shown which when clicked opens a browser window!

Is it possible to show a Peertube video (just like a Mastodon video) instead of a link to the video page?

There appears to be currently no working desktop client for following, commenting on and watching Peertube videos. There's [Cuttlefish](https://cuttlefish.ch/) but it appears to be abandoned.

### Implementation Details

- [ ] This should be an option in settings.
- [ ] This should be only available to some fediverse backends. (Include which ones on the above field).
- [ ] This is client-only (and shouldn't sync with the instance).
- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

GeopJr:
Thanks for the suggestion!

It will be done on #245 but since I don't really use peertube regularly, I might have missed special cases. Could you give it a try on accounts you usually watch?

You can either build it yourself or download one of the flatpak or snap artifacts when they are done https://github.com/GeopJr/Tuba/actions/runs/4956237749?pr=245

GeopJr:
Merged! If you notice anything, let me know!

RasheedAZ:
Sound started working after opening the volume slider and changing it. Continued working upon successive reopens. Thanks. 

No I hadn't changed the flatpak permissions.

Also, after clicking the back button, the sound of the video continued playing even when a different video was playing. However, after restarting the Tuba, I was not able to reproduce this either.

PR #245, opened 2023-05-12
feat(status): cards & peertube viewer

GeopJr:
Cards :sparkles: 

link:
image

video:
image

clicking on a card will show a dialog since it's an out-of-instance request and users should confirm for privacy reasons:
image

PeerTube :octopus: 

PeerTube videos should now play on the built-in media viewer

:information_source: qualities higher than 720p become laggy on my machine (gstreamer backend, I haven't tested on ffmpeg) so I set some limits (720, 480, 360 are preferred unless they are not available)

Similarly, PeerTube videos require out-of-instance requests and a similar dialog shows up:
image

fix: #244 

edit:

there's an option in settings to disable cards

### DIFF
--- a/data/dev.geopjr.Tuba.gschema.xml
+++ b/data/dev.geopjr.Tuba.gschema.xml
@@ -36,6 +36,9 @@
 		<key name="show-spoilers" type="b">
 			<default>false</default>
 		</key>
+		<key name="hide-preview-cards" type="b">
+			<default>false</default>
+		</key>
 		<key name="larger-font-size" type="b">
 			<default>false</default>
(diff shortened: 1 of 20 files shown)


GeopJr:
Ideally, inputstream + mediastream should be used for peertube but i've fallen into this bug: https://gitlab.gnome.org/GNOME/gtk/-/issues/4062

using file with uri works but doesn't seem to exit correctly (sound still plays even after destroying)

GeopJr:
The stopping issue seems to have been fixed my unpreparing the stream :shrug:

PR #253, opened 2023-05-15
feat(PreviewCard): support PeerTube mp4

GeopJr:
fix: https://github.com/GeopJr/Tuba/issues/244#issuecomment-1547708495

Two blockers:

- most instances DO NOT provide files, instead they rely on streaming mostly - the blender instance does provide files however (for testing)
- while seeking works when they are used, the video seems to stop updating after a while (audio still plays)

either will have to wait for https://github.com/GeopJr/Tuba/pull/245#issuecomment-1545345846 to be fixed and see if using inpustreams fixes the second point or try and figure out how to get Gtk.Video to use hlssink

I don't think we should create a custom video player.


GeopJr:
Handled by Clapper, might revisit in the very distant future if streaming inputstreams gets fixed upstream.

PR #254, opened 2023-05-16
Use github as provider for appdata-screenshots

fabiscafe:
Today someone was unable to build the AUR pkg of tuba. While testing this at work, I faces something similar. At my workplace imgur is blocked, so the appdata check at the end of the build process failed, because it's unable to reach the screenshots.

This PR changes the provider from imgur to githubusercontent.com, that is the official github domain. It's very unlikely that github and githubusercontent.com are blocked/not reachable at the same time. So the build should always work or fail for that single reason instead of 2 points or error.

I have tried to build tuba via my own github repo. This worked (even at work :clinking_glasses: ) and switched over to the original repo after that. Of course this will fail to build until the screenshots are merged.


GeopJr:
Thanks! I understand that the screenshots being hosted on an external service are causing issues but at the same time I don't think the repo is the best place to store them.

Tootle had them in the repo and looking at the sizes in history, they are the main reason cloning is ~5mb

image

Perhaps I should "host" them in another repo, git-lfs or another service maybe?

fabiscafe:
That's a question, hm. :sweat_smile:  An own repo, or branch could be an option. I sadly don't know about git-lfs. A different service would maybe hit the same problem, that it might not be reachable for the user who tries to build tuba.

GeopJr:
I switched to lfs on 5fb5d51ba60effde3e48d6b33c448be2b0326505

Thanks, merging!

PR #268, opened 2023-05-20
fix(bookwyrm): do not rely on title for bookwyrm detection

thiagojedi:
This was preventing preview on BW instances that changes the title (e.g. https://velhaestante.com.br/book/11624/s/the-name-of-the-wind ).

The regular expression used here is the same one used by BW itself (see [BOOK_PATH][bw-path], [SLUG][bw-slug], and [full path][bw-view]).

[bw-slug]: https://github.com/bookwyrm-social/bookwyrm/blob/ee1dd612fb223bf44f98026b8f42991cff97279f/bookwyrm/utils/regex.py#LL9C10-L9C34
[bw-path]: https://github.com/bookwyrm-social/bookwyrm/blob/ee1dd612fb223bf44f98026b8f42991cff97279f/bookwyrm/urls.py#L26
[bw-view]: https://github.com/bookwyrm-social/bookwyrm/blob/ee1dd612fb223bf44f98026b8f42991cff97279f/bookwyrm/urls.py#L649

### DIFF
--- a/src/API/Status/PreviewCard.vala
+++ b/src/API/Status/PreviewCard.vala
@@ -137,10 +137,9 @@ public class Tuba.API.PreviewCard : Entity {
 
 	public bool is_bookwyrm {
         get {
-			bool title_bw = title.last_index_of ("- BookWyrm") > 0;
-			bool url_bw = url.last_index_of ("/book/") > -1;
-
-			return kind == "link" && title_bw && url_bw;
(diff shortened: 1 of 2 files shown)


GeopJr:
LGTM, thanks!

I moved it to main so we don't create a new instance every time, passed the OPTIMIZE flag and handled the error

PR #287, opened 2023-05-26
feat: keyboard shortcut window

GeopJr:


GeopJr:
This is really not mobile friendly as far as I understand, especially when some actions have 4-5 sets of shortcuts, not sure what can be done.

PR #302, opened 2023-06-07
feat(Timeline): hide boosts and replies

GeopJr:
~~todo: figure out pagination on < page long content~~


GeopJr:
Not ideal way to do that.

PR #305, opened 2023-06-08
feat: remove tracking ids from links

GeopJr:
ported from https://source.puri.sm/Librem5/chatty/-/merge_requests/1229

draft until it's merged on Chatty

### DIFF
--- a/src/Utils/Tracking.vala
+++ b/src/Utils/Tracking.vala
@@ -52,8 +52,10 @@ public class Tuba.Tracking {
 
         var split_url = url.split_set ("?", 2);
         var search_terms = split_url[1].split_set ("&");
-
         var str = @"$(split_url[0])?";
+
+        var fragment_offset = split_url[1].last_index_of_char ('#');
+        var fragment = fragment_offset > -1 ? split_url[1].substring(fragment_offset) : "";
 
         foreach (var param in search_terms) {
             var not_tracking_id = true;
@@ -70,6 +72,6 @@ public class Tuba.Tracking {
             }
         }
 
-        return str.slice(0, -1);
+        return @"$(str.slice(0, -1))$fragment";
     }
 }
--- a/src/Utils/Tracking.vala
+++ b/src/Utils/Tracking.vala
@@ -1,0 +1,75 @@
+// Ported from Chatty
+// https://source.puri.sm/Librem5/chatty/-/merge_requests/1229
+
+public class Tuba.Tracking {
+    /* https://github.com/brave/brave-core/blob/5fcad3e35bac6fea795941fd8189a59d79d488bc/browser/net/brave_site_hacks_network_delegate_helper.cc#L29-L67 */
+    public const string[] tracking_ids = {
+        // Strip any utm_ based ones
+        "utm_",
+        // https://github.com/brave/brave-browser/issues/4239
+        "fbclid", "gclid", "msclkid", "mc_eid",
+        // https://github.com/brave/brave-browser/issues/9879
+        "dclid",
+        // https://github.com/brave/brave-browser/issues/13644
+        "oly_anon_id", "oly_enc_id",
+        // https://github.com/brave/brave-browser/issues/11579
+        "_openstat",
+        // https://github.com/brave/brave-browser/issues/11817
+        "vero_conv", "vero_id",
+        // https://github.com/brave/brave-browser/issues/13647
+        "wickedid",
+        // https://github.com/brave/brave-browser/issues/11578
+        "yclid",
+        // https://github.com/brave/brave-browser/issues/8975
+        "__s",
+        // https://github.com/brave/brave-browser/issues/17451
+        "rb_clickid",
+        // https://github.com/brave/brave-browser/issues/17452
+        "s_cid",
+        // https://github.com/brave/brave-browser/issues/17507
+        "ml_subscriber", "ml_subscriber_hash",
+        // https://github.com/brave/brave-browser/issues/18020
+        "twclid",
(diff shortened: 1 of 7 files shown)


src/Utils/Tracking.vala
```diff
@@ -0,0 +1,75 @@
...
+     * An URL takes the form: http://<host>:<port>/<path>?<searchpart>
+     *
+     */
+    public static string strip_utm (string url) {
+        if (!("?" in url)) return url;
+
+        var split_url = url.split_set ("?", 2);
+        var search_terms = split_url[1].split_set ("&");
+
+        var str = @"$(split_url[0])?";
```
bugaevc, line 56: Oh no no no no no, please don't parse URLs with regexps. Been there, done that, and it's horrible. Please use [`GUri`](https://docs.gtk.org/glib/struct.Uri.html)

GeopJr: It's not really parsing there:
`https://example.com/?test=1&foo=bar`
- split_url = { "https://example.com", "test=1&foo=bar" }
- search_terms = { "test=1", "foo=bar" }
- str = "https://example.com?" (first split_url item + `?`)

could that be abused?

bugaevc: > just to clarify that is not regex, it's vala's syntax for template literals

Yes, I figured that out right after I posted :|

> `https://example.com/?test=1&foo=bar`

That is a typical, but unfortunately very simplistic example of a URL. URLs can be _weird_. Special libraries for URL handling exist for a reason, otherwise everyone would just be doing `split('?')[1].split('&')`.

(I don't mean to imply that I'm any sort of expert on URLs! I'm terrified by URLs, and try to avoid parsing them myself whenever I can.)

Off the top of my head, some specific issues with this code:
* It doesn't handle `%`-encoding at all, does it?
* You do `str.slice(0, -1)`, but what if there were no params at all? -- then you'd strip the `?` away (e.g. `http://example.com/?`)
* You're checking `id in param`, but that would also fire on param (names and) values that only happen to contain the tracking id, such as `?good_param=i_hate_twclid`.
    * These are called _query params_, not _search terms_, by the way.
* You're not honoring the URL fragment, so the following `http://example.com/?utm_source=tracking#some_fragment` will have the `#some_fragment` stripped along with `?utm_source=tracking`.

On the other hand, the good things your implementation does (i.e. the mistakes it does not make):
* Does not reorder query params, such as e.g. by sorting or accidentally storing them in a map/dictionary that does not preserve the original order
* Preserves duplicate params, i.e. allows `?foo=bar&foo=baz`
* Does not choke on `?foo=bar=baz`
* Preserves value-less params, e.g. `?foo&bar=baz`, does not convert it into `?foo=&bar=baz`
* Does not mangle (decode, reencode, normalize) the params in any way
* Preserves leading and trailing `&`, such as in `?&foo=bar&`
* Does not really mess with the pre-query parts of the URL at all, so no risk of breaking that part

bugaevc: > but that would also fire on param (names and) values that only happen to contain the tracking id

As an actual example, consider this URL: https://www.google.com/search?q=what+is+twclid

It also would've been awesome to have unit tests for this all.

GeopJr: Thanks for the input! This is just a port from Chatty, since that's also not done, these should probably be raised there, but I guess I should deal with them here prior.

> It doesn't handle %-encoding at all, does it?

Should it? If a param key is url encoded then it's probably not part of the tracking ones

> You do str.slice(0, -1), but what if there were no params at all? -- then you'd strip the ? away (e.g. http://example.com/?)

That's the desired outcome. The link can either end up as
`http://example.com/?` if all parameters were tracking ones
or
`http://example.com/?foo=bar&` if there was at least one left

slicing will either turn it into `http://example.com/` or `http://example.com/?foo=bar`

> You're checking id in param, but that would also fire on param (names and) values that only happen to contain the tracking id, such as ?good_param=i_hate_twclid.
> These are called query params, not search terms, by the way.
> You're not honoring the URL fragment, so the following http://example.com/?utm_source=tracking#some_fragment will have the #some_fragment stripped along with ?utm_source=tracking.

I agree, I'm waiting for my purism account to be approved to raise these to upstream. But in the meantime I should tag @kop316 

(here's a real world example `http://www.example.com/store?tag=soft_clothes` (`oft_c`))

bugaevc: > That's the desired outcome. The link can either end up as `http://example.com/?` if all parameters were tracking ones

What I mean is, if the _original_ URL was `http://example.com/?` (obviously it doesn't contain any tracking params), you'd transform it into `http://example.com/`, which you should not do. *Hopefully* it won't matter to the server, but still.

GeopJr: Ok, it now uses Uri and Chatty's function as a fallback (with somewhat fragment support and now only checks against the key). Thanks for the input, I'll mark this as resolved, let me know if I missed anything!

bugaevc: Thank you!

> let me know if I missed anything!

The unit tests 😄 Without tests, we cannot be sure that this is working and not breaking valid, innocent URLs.

Basically each of my points (about preserving order, not deduplicating params, handling value-less params, and so on) could be a little unit test for this functionality. My Google search example and your "soft clothes" example, too.

I'm not familiar with testing in Vala; otherwise I'd contribute the tests myself.

GeopJr: Opened a separate issue as there are no unit tests at all right now #306

GeopJr:
Merging! I'll keep an eye out on the chatty MR in case they find any bugs I missed until they merge or close it

PR #311, opened 2023-06-15
fix: set composer dialog widgets to insensitive when publishing

VictoriaLacroix:
Currently when sending a post, the dialog does not change. This allows the user to accidentally press "Publish" or use one of the accelerators to send the post again. It's also just not good feedback on whether the post is sending.

This commit changes the dialog such that once the publish process begins, the "Publish" button, composer pages (edit/attach/poll), and on_commit () callback are all disabled until the commit transaction fails.

---

I didn't create an issue for this because I felt it would be a good opportunity to learn a bit of Vala/Adwaita if I tackled this myself, please let me know if you have feedback 🙏️

### DIFF
--- a/src/Dialogs/Composer/Dialog.vala
+++ b/src/Dialogs/Composer/Dialog.vala
@@ -16,6 +16,11 @@ public class Tuba.Dialogs.Compose : Adw.Window {
 		set { commit_button.add_css_class (value); }
 	}
 	public bool editing { get; set; default=false; }
+	private bool _working = false;
+	public bool working {
+		get { return _working; }
+		set { _working = value; update_commit_button (); update_pages (); }
+	}
 
 	ulong build_sigid;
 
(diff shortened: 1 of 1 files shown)


GeopJr:
Thanks for opening this PR! It looks good!

My only concern is that the commit button's sensitivity cases are (already) too complex and might be better to move it away from that. How about making the whole composer insensitive?

```vala
[GtkCallback] void on_commit () {
	this.sensitive = false;
	transaction.begin ((obj, res) => {
		try {
			transaction.end (res);
		} catch (Error e) {
			warning (e.message);
		} finally {
			this.sensitive = true;
		}
	});
}
```

VictoriaLacroix:
For some reason I had thought it'd be nice to be able to cancel a send, but seeing your comment I immediately realized that would not even be correct, so good call on that.

While I was there I also found a tiny two-line change I forgot to commit earlier, so I pushed that up as well.

src/Dialogs/Composer/Dialog.vala
```diff
@@ -87,9 +92,11 @@ public class Tuba.Dialogs.Compose : Adw.Window {
...
 	private void update_commit_button () {
 		var allow = false;
-		foreach (var page in t_pages) {
-			allow = allow || (page.can_publish && page.visible);
-			if (allow) break;
+		if (!working) {
+			foreach (var page in t_pages) {
+				allow = allow || (page.can_publish && page.visible);
+				if (allow) break;
+			}
```
GeopJr, line 99: This should probably be reverted for the reasons mentioned in the previous comment (commit button sensitivity complexity)

If the whole composer is insensitive then the commit button cannot be activated whether it itself is sensitive or not

src/Dialogs/Composer/Dialog.vala
```diff
@@ -190,14 +197,16 @@ public class Tuba.Dialogs.Compose : Adw.Window {
 	}
 
 	[GtkCallback] void on_commit () {
-		//working = true
+		if (working)
+			return; // avoid double send
```
GeopJr, line 201: I don't think on_commit can be called as long as the composer is insensitive but it doesn't hurt I guess!

I'll leave it up to you to decide whether to keep it or not, but if you don't then I'd suggest getting rid of `working` property altogether as it can be replaced with the snippet I suggested prior

VictoriaLacroix: Huh, yeah, turns out that on_commit is completely safe once the composer is insensitive, even when using the Ctrl+Return accelerator!

GeopJr:
LGTM, thank you so much!

Issue #316, opened 2023-06-18
[Bug]: There is no `Escape` from Editing


nekohayo:
### Describe the bug

The `Esc` keyboard shortcut is a very convenient standard way to dismiss dialogs or to go back. Currently it is only used for going back to previous screens, but it cannot dismiss the status/toot/post editing dialog, even if there are no unsaved changes.

It would be possible with a bunch of `Shift+Tab` actions to focus the Cancel button to escape that dialog with the keyboard, but much less convenient than having the `Esc` keyboard shortcut bound to it.

### Steps To Reproduce

1. Edit an existing post, or start editing a new one
2. Press `Esc` on the keyboard
3. Steel yourself as there is no Escape from New Post, [unless you are Snake Plissken](https://youtu.be/bqT09APfNKI?t=4).

### Instance Backend

Mastodon

### Operating System

Fedora 38

### Package

Flatpak

### Troubleshooting information

os: GNOME 44 (Flatpak runtime)
prefix: /app
flatpak: true
version: 0.3.2 (production)
gtk: 4.10.3 (4.10.3)
libadwaita: 1.3.2 (1.3.2)
libsoup: 3.4.2 (3.4.1)
libgtksourceview: 5.8.0 (5.8.0)

GeopJr:
Last time this got brought up the decision was to tie Escape's closing function to whether you are able to commit the post - Esc works only if you cannot commit a post - for the sake of accidentally losing progress. How about a dialog? That would allow us to add a "Save as Draft" button in the future too

![image](https://github.com/GeopJr/Tuba/assets/18014039/ceb2cfdf-8723-4fd8-b95a-1feb6f1943b1)

nekohayo:
If there are unsaved modifications I wouldn't mind the confirmation dialog indeed, with the future opportunity to save a draft.

I think having the Esc shortcut available or not available depending on some invisible condition would not be as good as an approach, in comparison; it would be unpredictable and confusing.

Issue #263, opened 2023-05-20
[Request]: Ask the user for confirmation when closing a non-empty editor window


LukaszH77:
### Describe the request

Tuba doesn't ask the user if they are sure they want to close the editor w/o sending an already written post, and I've lost several posts because of that and my own carelessness :)

### Implementation Details

- [ ] This should be an option in settings.
- [ ] This should be only available to some fediverse backends. (Include which ones on the above field).
- [X] This is client-only (and shouldn't sync with the instance).
- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

nekohayo:
I believe a better way would be to autosave the contents and to restore them the next time you invoke the composer (or maybe show an InfoBar with a preview of the old contents and offer to restore them), rather than to prompt for the "are you sure".

The autosave approach would also have the advantage of offering protection against outages (network or power), computer crashes, the user needing to quit the app/logout for some reason, etc.

LukaszH77:
I like your idea. Phanpy does it like that and it works well. 

![Zrzut ekranu z 2023-05-20 23-04-51](https://github.com/GeopJr/Tuba/assets/121253127/11211f25-dfc1-4b73-bdb3-6bbfccfc4c84)

PR #317, opened 2023-06-19
feat(Composer): show dialog on exit

GeopJr:
fix: #316 ?
fix: #263 ? (I just noticed this issue)
fix: #319 

~~todo:~~

- ~~finish alt text editing~~
- ~~add alt text changed on equal ()~~

### DIFF
--- a/data/ui/dialogs/compose.ui
+++ b/data/ui/dialogs/compose.ui
@@ -27,7 +27,8 @@
             <child type="start">
               <object class="GtkButton" id="close_button">
                 <property name="label" translatable="yes">Cancel</property>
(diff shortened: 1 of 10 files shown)


LukaszH77:
Tested 5437dd3 for a bit, and it crashes every time I try to edit a post with image w/o an alt text. There's nothing in the logs after the crash, GET source is the last line.

GeopJr:
It should be done now, thanks!

#319 required another big change, if you notice anything let me know!

LukaszH77:
Tested it for a bit yesterday and today and so far so good.

PR #329, opened 2023-06-26
feat(Status): pinning

GeopJr:
fix: #328 

TODO:

- probably lock it only to self profile
- rethink if removing from cache is the best way to update it with refresh


LukaszH77:
Looks like it works OK. But it shouldn't be available for Pixelfed accounts, because that platform doesn't support pinning posts.

GeopJr:
There is no backend sniffing at all right now. My original thought was not to do it at all in an attempt to get people to either update or for fedi server developers to test Mastodon compat, but the closer we get to Misskey API support, the more it is needed (since we need a way to check if Misskey compat should be enabled or not). For now, all backends are treated the same:
Pleroma, Akkoma have the Explore tab even if they don't support it.
GoToSocial has polls & "edit" in the context menu even if it's not implemented.
Mastodon < 4.0.0 has hashtag tracking even if it wasn't implemented until 4.0.0
etc.

PR #339, opened 2023-06-30
fix: add .flatpak-builder/ directory to .gitignore

doasu:
When building the app with `flatpak-builder`, a `.flatpak-builder/` directory is created for the build environment.

_Since (afaik) build directories/environments shouldn't be pushed to repositories, it would be logical to add this exception._

<sub>This is my first pull request, please let me know if I did anything wrong; any and all criticism is very much appreciated!</sub>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ subprojects/packagecache
 *~
 builddir
 *.snap
+.flatpak-builder


GeopJr:
LGTM, thanks!

PR #344, opened 2023-07-02
fix(Status): name label styling

GeopJr:
1. headings (x boosted / favourited / etc) were some pixels to the right (button + padding + box spacing) but now should match the name label, label being in the same column as the status content
2. name labels with emojis were some pixels higher than the handle due to the emoji. Fixing it is a bit hacky but it works: increase line height to 1.8 and negative padding vertically to decrease the size of the button


GeopJr:
:/ when the name wraps the line height is too much, maybe lowering the emoji pixel size?

PR #353, opened 2023-07-05
feat: zoom in on custom emoji hover

GeopJr:
[Screencast from 2023-07-05 16-51-20.webm](https://github.com/GeopJr/Tuba/assets/18014039/464e2e37-a73c-453a-83a3-671770e30c30)

Custom emojis are hard to distinguish sometimes due to their size / being too pixelated. This PR adds the PleromaFE feature that scales them up on hover


GeopJr:
Should this be enabled by default? (it's a setting users can toggle)

cc: @bertob

PR #365, opened 2023-07-10
feat(Profile): editing

GeopJr:
fix: #364 

TODO:

- ~~test on *oma~~
- ~~test > 4 fields on glitch~~
- ~~check if new saving format breaks already saved keys~~
- ~~test errors (large files etc)~~
- ~~save on update_credentials cb~~
- ~~remove self from entity cache?~~


GeopJr:
document-edit-symbolic overuse but it'll do for now :/

PR #368, opened 2023-07-11
chore(flatpak): do not use my fork of gspell-4

GeopJr:


GeopJr:
nvm, looks like it's still needed.

PR #371, opened 2023-07-12
feat: libspelling support

GeopJr:
gspell-4 is still available as a fallback

The only difference I noticed (apart from some UI) is that your cursor needs to be on the word you are looking suggestions up for.

cursor next to word:
image

cursor in word:
image

:warning: If GtkSourceView is missing, then only gspell-4 is available


GeopJr:
cc: @LukaszH77

LukaszH77:
Whoa, that was quick!

PR #379, opened 2023-07-14
fix: Escape markup in poll options

majormajors:
Poll option titles can currently be interpreted as html, causing them to be displayed incorrectly at times. This is inconsistent with other Mastodon/ActivityPub clients. All documentation on how poll options should be formatted also seems to point to them only ever being plain text.

This small change escapes characters that can be interpreted as markup when rendering polls.

Before:
screenshot_2023-07-14T09:13:48

After:
screenshot_2023-07-14T09:11:44

### DIFF
--- a/src/Widgets/VoteBox.vala
+++ b/src/Widgets/VoteBox.vala
@@ -1,6 +1,7 @@
 using Gtk;
 using Gdk;
 using Gee;
+using GLib;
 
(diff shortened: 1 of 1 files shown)


GeopJr:
Nice catch & thanks! Instead of escaping it, I disabled markup instead

PR #405, opened 2023-07-22
Shiny new widgets \o/

bugaevc:
Hi,

this is an attempt to port Tuba to the new libadwaita widgetry ([click](https://blogs.gnome.org/alicem/2023/06/15/rethinking-adaptivity/), [click](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/migrating-to-breakpoints.html)). I'm not an expert in GTK, nor in Tuba's codebase, so I most likely messed something up — please let me know!

This PR only touches the main window (and not even all of its views?), not the compose dialog.

Screenshot from 2023-07-22 22-13-46
Screenshot from 2023-07-22 23-55-08
Screenshot from 2023-07-22 23-55-59
Screenshot from 2023-07-22 23-56-53

### DIFF
--- a/src/Dialogs/MainWindow.vala
+++ b/src/Dialogs/MainWindow.vala
@@ -1,7 +1,7 @@
 [GtkTemplate (ui = "/dev/geopjr/Tuba/ui/dialogs/main.ui")]
 public class Tuba.Dialogs.MainWindow: Adw.ApplicationWindow, Saveable {
-	[GtkChild] public unowned Adw.Flap flap;
-	[GtkChild] unowned Adw.Leaflet leaflet;
+	[GtkChild] unowned Adw.NavigationView navigation_view;
+	[GtkChild] public unowned Adw.OverlaySplitView split_view;
 	[GtkChild] unowned Views.Sidebar sidebar;
 	[GtkChild] unowned Gtk.Stack main_stack;
 	[GtkChild] unowned Views.MediaViewer media_viewer;
@@ -14,7 +14,7 @@ public class Tuba.Dialogs.MainWindow: Adw.ApplicationWindow, Saveable {
 		var gtk_settings = Gtk.Settings.get_default ();
 	}
 
-	private Views.Base main_base;
+	private Adw.NavigationPage main_page;
 	public MainWindow (Adw.Application app) {
 		Object (
 			application: app,
@@ -23,8 +23,8 @@ public class Tuba.Dialogs.MainWindow: Adw.ApplicationWindow, Saveable {
 			resizable: true
 		);
 		sidebar.set_sidebar_selected_item (0);
-		main_base = new Views.Main ();
-		open_view (main_base);
+		main_page = new Adw.NavigationPage (new Views.Main (), "Main");
+		navigation_view.add (main_page);
 
 		if (Build.PROFILE == "development") {
 			this.add_css_class ("devel");
@@ -121,7 +121,7 @@ public class Tuba.Dialogs.MainWindow: Adw.ApplicationWindow, Saveable {
 
 	public Views.Base open_view (Views.Base view) {
 		if (
-			leaflet.visible_child == view
+			navigation_view?.visible_page?.child == view
 			|| (
 				last_view != null
 				&& last_view.label == view.label
@@ -129,13 +129,13 @@ public class Tuba.Dialogs.MainWindow: Adw.ApplicationWindow, Saveable {
 			)
 		) return view;
 
-		if (last_view != null && !last_view.is_main && view.is_sidebar_item) {
-			leaflet.insert_child_after (view, main_base);
+		Adw.NavigationPage page = new Adw.NavigationPage (view, view.label);
+		if (view.is_sidebar_item && navigation_view.visible_page != main_page) {
+			navigation_view.replace ({ main_page, page });
 		} else {
-			leaflet.append (view);
+			navigation_view.push (page);
 		}
 
-		leaflet.visible_child = view;
 		return view;
 	}
 
@@ -147,19 +147,19 @@ public class Tuba.Dialogs.MainWindow: Adw.ApplicationWindow, Saveable {
 
 		if (last_view == null) return true;
 
-		leaflet.navigate (Adw.NavigationDirection.BACK);
+		navigation_view.pop ();
 		return true;
 	}
 
 	public void go_back_to_start () {
 		var navigated = true;
 		while (navigated) {
-			navigated = leaflet.navigate (Adw.NavigationDirection.BACK);
+			navigated = navigation_view.pop ();
 		}
 	}
 
 	public void scroll_view_page (bool up = false) {
-		var c_view = leaflet.visible_child as Views.Base;
+		var c_view = navigation_view.visible_page.child as Views.Base;
 		if (c_view != null) {
 			c_view.scroll_page (up);
 		}
@@ -174,9 +174,8 @@ public class Tuba.Dialogs.MainWindow: Adw.ApplicationWindow, Saveable {
 	public void switch_timeline (int32 num) {}
 
 	[GtkCallback]
-	void on_view_changed () {
-		var view = leaflet.visible_child as Views.Base;
-		on_child_transition ();
+	void on_visible_page_changed () {
+		var view = navigation_view.visible_page.child as Views.Base;
 
 		if (view.is_main)
 			sidebar.set_sidebar_selected_item (0);
@@ -193,17 +192,4 @@ public class Tuba.Dialogs.MainWindow: Adw.ApplicationWindow, Saveable {
 
 		last_view = view;
 	}
-
-	[GtkCallback]
-	void on_child_transition () {
-		if (leaflet.child_transition_running)
-			return;
-
-		Gtk.Widget unused_child = null;
-		while ((unused_child = leaflet.get_adjacent_child (Adw.NavigationDirection.FORWARD)) != null) {
-			leaflet.remove (unused_child);
-			unused_child.dispose ();
-		}
-	}
-
 }
--- a/src/Dialogs/ProfileEdit.vala
+++ b/src/Dialogs/ProfileEdit.vala
@@ -1,12 +1,12 @@
 [GtkTemplate (ui = "/dev/geopjr/Tuba/ui/dialogs/profile_edit.ui")]
 public class Tuba.Dialogs.ProfileEdit : Adw.Window {
 	~ProfileEdit () {
-		message (@"Destroying ProfileEdit for $(profile.handle)");
+		debug (@"Destroying ProfileEdit for $(profile.handle)");
 	}
 
 	public class Field : Adw.ExpanderRow {
 		~Field () {
-			message ("Destroying ProfileEdit.Field");
+			debug ("Destroying ProfileEdit.Field");
 		}
 
 		Adw.EntryRow key_row;
--- a/src/Views/Sidebar.vala
+++ b/src/Views/Sidebar.vala
@@ -1,5 +1,5 @@
 [GtkTemplate (ui = "/dev/geopjr/Tuba/ui/views/sidebar/view.ui")]
-public class Tuba.Views.Sidebar : Gtk.Box, AccountHolder {
+public class Tuba.Views.Sidebar : Gtk.Widget, AccountHolder {
 
 	[GtkChild] unowned Gtk.ToggleButton accounts_button;
 	[GtkChild] unowned Gtk.Stack mode;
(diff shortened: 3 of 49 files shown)


GeopJr:
Thank you so much for this PR!

src/Dialogs/MainWindow.vala
```diff
@@ -125,22 +125,8 @@ public class Tuba.Dialogs.MainWindow: Adw.ApplicationWindow, Saveable {
...
-			)
-		) return view;
-
-		if (last_view != null && !last_view.is_main && view.is_sidebar_item) {
-			leaflet.insert_child_after (view, main_base);
-		} else {
-			leaflet.append (view);
-		}
-
-		leaflet.visible_child = view;
```
GeopJr, line 143: This part exists for two reasons:
1. Avoiding opening the current view again (example: if you are on "Lists" and click "Lists" again it won't open it again)
2. If you are in a sidebar view and select another it removes all other pages and then adds it (or rather View -> Home -> Another View) (removal happens elsewhere), this is to avoid circular navigation (Home -> Lists -> Hashtags -> Lists becomes Home -> Lists). Oh and for performance reasons (many of the views will keep an active websocket connection as long as they are not disposed). This has its cons but it's probably a discussion to be held in a different issue

src/Dialogs/MainWindow.vala
```diff
@@ -198,17 +183,4 @@ public class Tuba.Dialogs.MainWindow: Adw.ApplicationWindow, Saveable {
...
-		if (leaflet.child_transition_running)
-			return;
-
-		Widget unused_child = null;
-		while ((unused_child = leaflet.get_adjacent_child (Adw.NavigationDirection.FORWARD)) != null) {
-			leaflet.remove (unused_child);
-			unused_child.dispose ();
-		}
-	}
-
```
GeopJr, line 213: Memory leak

That callback would remove pages from the leaflet when you navigated backwards and then dispose them. For AdwNavigationView, the replacement would be a combination of Adw.NavigationView#popped(?) and Adw.NavigationPage#hidden. It'll be a bit tricky to implement as it first needs to be removed from the NavigationView (Adw.NavigationView#remove) before it gets disposed but only when Adw.NavigationPage#hidden is emitted (otherwise the transition won't play and the page will just disappear) (there might be better ways to do this than that however)

To check if something is memory leaking at a glance, look at the logs; going back *should* log a "Destroying <widget>" log for the view and all its children (e.g. in the current version of Tuba, opening the federated view and then going backwards should display a "destroying status" log for every post and some for the view

src/Dialogs/MainWindow.vala
bugaevc: I see, thanks! I'll try to restore this logic then

src/Dialogs/MainWindow.vala
bugaevc: I must be missing something fundamental here: wouldn't the navigation view pop (and so unreference) the view by itself when you go back? And then if nothing else references the view it gets destroyed. No?

src/Dialogs/ProfileEdit.vala
```diff
@@ -415,7 +415,7 @@ public class Tuba.Dialogs.ProfileEdit : Adw.Window {
 
 			chooser.response.connect (id => {
 				switch (id) {
-					case ResponseType.ACCEPT:
+					case Gtk.ResponseType.ACCEPT:
```
bugaevc, line 418: No idea why it was working before and not working now to omit `Gtk.` here

src/Views/Sidebar.vala
```diff
@@ -59,6 +47,10 @@ public class Tuba.Views.Sidebar : Box, AccountHolder {
 			}
 	};
 
+	static construct {
+		typeof (Widgets.EmojiLabel).ensure ();
+	}
```
bugaevc, line 53: Nor why this suddenly became required

src/Dialogs/MainWindow.vala
bugaevc: But indeed you're right and I see that the destructor doesn't get called. So is something else still referencing the view?

src/Dialogs/MainWindow.vala
bugaevc: Pushed an update that tries to use `navigation_view.replace ()` for sidebar item pages. This is still not perfect, because now the Home row in the sidebar doesn't seem to have any effect at all. Also possibly this should be doing `navigation_view.replace ({ home_page, page });` and not just `{ page }`, to enable going back home from other sidebar views; and then also if we're already at `home`, we still should be doing `push ()`, not `replace ()`, so we get the animation (& building the intuition that it's possible to go back).

Please tell me what you think!

GeopJr: I'm not that fond of replacing the whole stack every time for the reasons you mentioned, how about this:

```diff
diff --git a/src/Dialogs/MainWindow.vala b/src/Dialogs/MainWindow.vala
index fabce38..ca31ca1 100644
--- a/src/Dialogs/MainWindow.vala
+++ b/src/Dialogs/MainWindow.vala
@@ -19,7 +19,6 @@ public class Tuba.Dialogs.MainWindow: Adw.ApplicationWindow, Saveable {
 		var gtk_settings = Gtk.Settings.get_default ();
 	}
 
-	private Views.Base main_base;
 	public MainWindow (Adw.Application app) {
 		Object (
 			application: app,
@@ -28,8 +27,7 @@ public class Tuba.Dialogs.MainWindow: Adw.ApplicationWindow, Saveable {
 			resizable: true
 		);
 		sidebar.set_sidebar_selected_item (0);
-		main_base = new Views.Main ();
-		open_view (main_base);
+		open_view (new Views.Main (), "main");
 
 		if (Build.PROFILE == "development") {
 			this.add_css_class ("devel");
@@ -124,13 +122,29 @@ public class Tuba.Dialogs.MainWindow: Adw.ApplicationWindow, Saveable {
 		}
 	}
 
-	public Views.Base open_view (Views.Base view) {
-		var page = new Adw.NavigationPage (view, view.label);
-		if (view.is_sidebar_item) {
-			navigation_view.replace ({ page });
+	public Views.Base open_view (Views.Base view, string? tag = null) {
+		if (
+			navigation_view?.visible_page?.child == view
+			|| (
+				last_view != null
+				&& last_view.label == view.label
+				&& !view.allow_nesting
+			)
+		) return view;
+
+		Adw.NavigationPage page;
+		if (tag == null) {
+			page = new Adw.NavigationPage (view, view.label);
 		} else {
-			navigation_view.push (page);
+			page = new Adw.NavigationPage.with_tag (view, view.label, tag);
 		}
+
+		if (last_view != null && !last_view.is_main && view.is_sidebar_item) {
+			navigation_view.pop_to_tag ("main");
+		}
+
+		navigation_view.push (page);
+
 		return view;
 	}
 

```

It goes back to the page tagged as "main" and then pushes the new page

src/Dialogs/MainWindow.vala
GeopJr: > I must be missing something fundamental here: wouldn't the navigation view pop (and so unreference) the view by itself when you go back? And then if nothing else references the view it gets destroyed. No?

It *should*, but memory leaks occur for random reasons sometimes in Vala (https://github.com/search?q=repo%3AGeopJr%2FTuba+memory+leak&type=commits (especially with lambdas))

> So is something else still referencing the view?

That was added pre-Tuba likely for a similar reason (removing it from the leaflet should be disposing it too). Investigating it is probably not worth doing on this PR

src/Dialogs/ProfileEdit.vala
GeopJr: My bad! Copy-pasted the GTK < 4.10 deprecation compat without checking it, unsure why the < 4.10 was used considering adw 1.4 requires 4.11 :/ (either way it will be removed on when adw 1.4)

src/Views/Sidebar.vala
GeopJr: Probably a `master` issue, not worth investigating yet

src/Dialogs/MainWindow.vala
bugaevc: Pushed an implementation, please try it out.

src/Dialogs/MainWindow.vala
bugaevc: > memory leaks occur for random reasons sometimes in Vala (especially with lambdas))

Oh my, this is very very bad. And you're working around this by calling `g_object_run_dispose ()` when popping a page? You should definitely investigate where the leak/refcycle is coming from and find a proper way to prevent it, rather than doing it like this. But you've got a point about not doing this in this PR.

src/Dialogs/MainWindow.vala
GeopJr: Looks good!

src/Dialogs/MainWindow.vala
bugaevc: OK, I pushed an attempt to do something like that. The navigation view emits `popped` _before_ actually unparenting the widget, so it doesn't work to `dispoe ()` it at that point. One would think that running `dispose ()` in the net iteration of the loop (`g_idle_add_once ()`) is the obvious solution, but it seems that `Idle.add_once ()` is even more broken/cursed in Vala: it causes things to segfault immediately, and the generated C code makes no sense.

So what I did that seems to work is in the widget itself, watch for it getting unparented, and run `dispose ()` at that point.

GeopJr: Sounds good and looks much more elegant! Thanks!

bertob:
What's the status here? Would be sweet to have this in the next release, with GNOME 45 coming out next month :)

bugaevc:
This is good to go from my side, except I'd probably have to rebase, and use the 45 runtime instead of master. @GeopJr probably needs to do another round of review and testing.

GeopJr:
This is not stale but I have postponed it until the cleanup PR is done!

oscfdezdz:
Hi, while using this branch I noticed some deprecation warnings and that the preferences dialog, could use other new libadwaita widgets. I've made the changes in [this fork](https://github.com/oscfdezdz/Tuba/tree/shiny-new-widgets) and after testing it, it seems that everything works fine although it never hurts to test it more, especially the one that replaces AdwLeaflet in the new account dialog. Feel free to cherry-pick the commits or I can open a PR against this branch if you want 😄 

Captura desde 2023-08-22 13-50-30
Captura desde 2023-08-22 13-50-42
Captura desde 2023-08-22 13-50-51
Captura desde 2023-08-22 13-51-03

GeopJr:
Main is now somewhat ready, could you rebase?

There have been many changes in that cleanup PR, most importantly for this PR however: All ListBoxes became ListViews. When in a scrolledwindow they __have to__ be in the following structure: GtkScrolledWindow -> AdwClampScrollable -> ListView. No boxes or other children as it will cause issues with the ListView's size.

That might need some changes in anything that inherits from the Base widget

(I've kind-of rushed the cleanup PR a bit since I didn't want to make rebasing even more difficult so there might be bugs!)

If you are busy, let me know and I'll rebase it instead!

bugaevc:
Yay! Rebased (that took some work) & added @oscfdezdz's changes. Since `column_view` is gone, I had to improvise again, and this time placed the spinner into the overlay. It still seems to work fine though 🙂

Please review & test!

bertob:
Just built the branch and gave it a spin, a few things I noticed (not all necessarily coming from this MR, haven't tested the app in a while):

- The body text in the profile descriptions is bold

image

- The focus border on posts should have the same shape and consistent padding all around (currently there's extra vertical padding):

image

- The focus border on boost profile pictures should be circular

image

- I'd drop the padding between cards on narrow views, and just have simple divider lines between posts:

image


- Could also be done as a followup, but I'd merge this view switcher into the sidebar at large sizes

image

GeopJr:
> Yay! Rebased (that took some work) & added [at]oscfdezdz's changes. Since `column_view` is gone, I had to improvise again, and this time placed the spinner into the overlay. It still seems to work fine though 🙂
> 
> Please review & test!

Thanks and sorry about that 🙇

Everything looks good, I'll go ahead and merge it! Thank you so much for making this PR!




> Just built the branch and gave it a spin, a few things I noticed (not all necessarily coming from this MR, haven't tested the app in a while):
> 
>     * The body text in the profile descriptions is bold
>     * The focus border on posts should have the same shape and consistent padding all around (currently there's extra vertical padding):
>     * The focus border on boost profile pictures should be circular

Nice catch! I'll get on it on a different PR!

>     * I'd drop the padding between cards on narrow views, and just have simple divider lines between posts:
>     * Could also be done as a followup, but I'd merge this view switcher into the sidebar at large sizes

This PR is only the start of #228, as long as they are in the mockup (they are), we'll get there!

Issue #313, opened 2023-06-18
[Bug]: Lag when resizing text content


orowith2os:
### Describe the bug

When resizing the window, specifically when resizing text content (image is fine), the resize will lag behind a *lot*.

Maybe an issue with libadwaita, but considering I have less issues with other text-based libadwaita applications, I'm thinking it's a Tuba issue. Other applications, when faced with tons of resizing text, are a lot smoother.

### Steps To Reproduce

1. Log into Tuba
2. Make the window small
3. Find some text content
3. Resize the window such that the text content adjusts itself a lot

### Logs and/or Screenshots

```
❯ dev.geopjr.Tuba 

(dev.geopjr.Tuba:2): Gtk-WARNING **: 03:14:13.466: Unknown key gtk-modules in /var/home/oro/.var/app/dev.geopjr.Tuba/config/gtk-4.0/settings.ini
Tuba-Message: 03:14:13.539: Application.vala:114: os: GNOME 44 (Flatpak runtime)
Tuba-Message: 03:14:13.539: Application.vala:114: prefix: /app
Tuba-Message: 03:14:13.539: Application.vala:114: flatpak: true
Tuba-Message: 03:14:13.539: Application.vala:114: version: 0.3.2 (production)
Tuba-Message: 03:14:13.539: Application.vala:114: gtk: 4.10.3 (4.10.3)
Tuba-Message: 03:14:13.539: Application.vala:114: libadwaita: 1.3.2 (1.3.2)
Tuba-Message: 03:14:13.539: Application.vala:114: libsoup: 3.4.2 (3.4.1)
Tuba-Message: 03:14:13.539: Application.vala:114: libgtksourceview: 5.8.0 (5.8.0)
Tuba-Message: 03:14:13.541: SecretAccountStore.vala:11: Using libsecret v0.20.5
```

All other logs are irrelevant, and nothing shows when resizing.

### Instance Backend

Mastodon

### Operating System

Fedora Silverblue 38

### Package

Flatpak

### Troubleshooting information

os: GNOME 44 (Flatpak runtime)
prefix: /app
flatpak: true
version: 0.3.2 (production)
gtk: 4.10.3 (4.10.3)
libadwaita: 1.3.2 (1.3.2)
libsoup: 3.4.2 (3.4.1)
libgtksourceview: 5.8.0 (5.8.0)

### Additional Context

I can also reproduce the lag when clicking on my profile and resizing that.

GeopJr:
[...]

Issue #179, opened 2023-04-10
[Request]: Migrate timelines to ListView


bleakgrey:
### Describe the request

This is going to be a long one.

As discussed previously, the status widget in its current state is causing us quite a bit of trouble. It's being pestered by 3 major issues:

**Property binding issues**
There are some obscure errors caused by property binding lifecycle management, which renders some things broken (#103).

I believe this can be addressed by using Expressions introduced in GTK4. I'll quote the [GTK team's post](https://blog.gtk.org/2020/06/07/scalable-lists-in-gtk-4/) directly:
`GObject’s GBinding mechanism goes in the right direction, but is not flexible enough to handle situations where you may need to bind properties of sub-objects or widgets deep inside the hierarchy of your widget, and where the objects in questions may not even exist at the time you are setting up the binding.`

I believe this is something we should try using instead.

 
   
**View architecture**
Back when Tootle was a thing, I used .ui files to describe views. I found it a good way to separate layout from logic, and I still think we should stick to it. However, Tuba widened its scope quite a bit by adding support for new instance backends and instance features. I proposed a potential solution in #57 to accommodate these changes, but now I'm rather skeptical about my proposal as:
a) the performance gains in #128 turned out to be rather negligent
b) it seems to make the developer's quality of life worse by reverting back to defining views via code

**Scalability**
Due to the fact that timelines use a ListBox (which constructs a widget for each status) _and_ continuous scrolling, it may be a bottleneck for some low-end devices. Timelines can be modernized by porting them to use a ListView instead, which [GTK devs recommend doing](https://docs.gtk.org/gtk4/section-list-widget.html#behind-the-scenes) anyway:
`While it is not a problem for short lists to instantiate widgets for every item in the model, once lists grow to thousands or millions of elements, this gets less feasible. Because of this, the views only create a limited amount of listitems and recycle them by binding them to new items. In general, views try to keep listitems available only for the items that can actually be seen on screen.`

All things considered, I think we've been optimizing the wrong thing. I've been searching for a way to reduce the complexity of the widget's hierarchy, when it makes much more sense to reduce the amount of the status widgets visible.

I propose doing the following:
- Make the status widget use containers bound to `ListModel`s when possible, especially so for emoji reactions and attachments. If a container's data model is empty, it's just an empty container that just sits in there and doesn't need much. However, it allows for seamless live updates via, say, stream events if such functionality should be needed.
[...]

Issue #57, opened 2023-01-13
[Feature request]: Status widget needs to be refactored


bleakgrey:
### Describe the bug

The more I think about `Widgets.Status`, the more I agree that its architecture has become questionable.

Currently, each `Widgets.Status` has a pre-instanced `Widgets.VoteBox`, `Widgets.AttachmentBox`, and emoji `Gtk.FlowBox`. These widgets stay even if their status entity doesn't really need them, so they just hang there being invisible. There really is no need to create so many useless widgets.

Ideally, the entire content box should be generated dynamically on demand (when bound to a `API.Status`) to avoid creating a larger memory footprint. Some widgets can be skipped altogether if the account backend doesn't support some features (like emoji reactions).

To make things worse, Notification widgets derive from `Widget.Status`, so they all succumb to this behavior as well (even though it makes no sense for a follow request to have a `VoteBox` at its disposal).

### Steps To Reproduce

Inspect https://github.com/GeopJr/Tooth/blob/main/data/ui/widgets/status.ui#L220-L226 and https://github.com/GeopJr/Tooth/blob/main/data/ui/widgets/status.ui#L257-L264

### Instance Backend

Mastodon

### Operating System

Pop!_OS 22.04 LTS

### Package

Flatpak

### Troubleshooting information

flatpak: false
version: main-15b8f4a (development)
gtk: 4.8.4 (4.8.4)
libadwaita: 1.2.1 (1.2.1)
libsoup: 2.74.2 (2.74.2)

GeopJr:
Definitely agree!

I do have one question for Status widgets: Why do we even bind all their properties? Do they (most of them at least) actually change?

Constructing them once (with only the children they need) is obviously the desired solution but if e.g. they can get a spoiler at some point then I doubt the memory saving is worth the cost of destroying the content label -> creating a Stack, pages, spoiler button, spoiler label, icons, content label...

bleakgrey:
Only a handful of properties need to be reactive, like `reblogged`, `favourited`, and `bookmarked`. For instance, if you bookmark a status in a Conversation, it becomes bookmarked in other Timelines as well for consistency.

If Mastodon supported editing statuses, it would make sense to make `content` reactive too. I'm not sure about other backends though.

bleakgrey:
Here's a full list of properties that definitely should be reactive:
- `reblogged`
- `favourited`
- `bookmarked`
- `replies_count`
- `reblogs_count`
- `favourites_count`

[...]

PR #424, opened 2023-08-02
feat: cleanup

GeopJr:
I want to go through as much as I can and clean up, optimize & simplify anything I find

Especially:

- ~~Move complicated UIs to Builder files~~
- Convert Builder files to Blueprint since they are getting too complicated
- Remove as many lambdas as possible
- ~~Remove widgets that are obsolete~~
- ~~Avoid abusing bindings and signals~~
- Simplify views, functions, widgets etc
- Make sure nothing is leaking
- Fix any potential segfaults
- Remove unused options
- Reorganize options
- Docs

fix: #57 
close: #128
fix: #179
fix: #313 (?)


GeopJr:
ListView needs some fixes still:

- Sometimes start and end are cropped
- Scroll to doesnt always scroll
- Some widgets do not open
- STYLES STYLES STYLES

With ListViews we no longer have access to the widgets so we have to think of other ways to dynamically modify them - like for thread lines, it has to be saved in API.Status but only be visible in threads and when it's not expanded.

GeopJr:
Profiles are a bit weird. I have to use the header factory for the cover however it doesn't play well with pull to refresh :/

GeopJr:
I need to merge this as is and continue on the rest as individual PRs as I need the new widgets to continue.

If you are a user, the nightlies from now until GNOME 45 will depend on the master runtime.

PR #434, opened 2023-08-06
experiment-feat(build-aux): use clang

GeopJr:


GeopJr:
Nope, did not improve compilation times on aarch64

PR #440, opened 2023-08-08
feat(notifications): actions

GeopJr:
On mention:
- Reply

On follow:
- Remove from Followers
- Follow Back

On follow request:
- Decline
- Accept

On click:
- Open


GeopJr:
Should probably embed the id of the account that received the notification instead of not showing the buttons

PR #465, opened 2023-08-31
data: Fix translation related issues

sabriunal:
**data: Remove non-exist file from POTFILES** 
- Remove LockableToogleButton.vala from the POTFILES. This file seems not exist and blocks the POT file updates.

**data: update the POTFILES**
- Order entries alphabetically respecting folder structures.
- Remove unused files.


GeopJr:
Thanks for this PR!

Would you mind somewhat automating it instead of keeping track of every file manually?
Something like: `find -iname "*.ui" -o -iname "*.vala" -o -iname "*.in" | sort`

sabriunal:
You can create a bash script (sh) file on the main directory of your project containing this entries

```bash
#!/bin/bash

find ./ -type f -name "*.in" | sort > po/POTFILES
echo "" >> po/POTFILES
find ./ -type f -name "*.ui" -exec grep -l "translatable=\"yes\"" {} \; | sort >> po/POTFILES
echo "" >> po/POTFILES
find ./ -type f -name "*.vala" -exec grep -l "_(\"" {} \; | sort >> po/POTFILES

```

GeopJr:
LGTM thanks!

I added the script and ran it

PR #469, opened 2023-08-31
data: Update Help Overlay

sabriunal:
- Fix overflowing the Go Back entry
- Add missing shortcuts: Close Window, About, Scroll Up, Scroll Down

### DIFF
--- a/data/gtk/help-overlay.ui
+++ b/data/gtk/help-overlay.ui
@@ -27,11 +27,10 @@
                 <property name="action-name">app.search</property>
               </object>
             </child>
-            <!-- Too many, causes overflow -->
             <child>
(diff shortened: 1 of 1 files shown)


GeopJr:
Thanks for this PR!

Ideally, the scroll actions would be under a "Timelines" category but that wouldn't solve the overflowing problem.

Do you think we should list all Go Back options? (As different entries)

sabriunal:
Two shortcuts for an action enough for users.  Actuall, the Escape key is
well known for this task.

Evan Paterakis ***@***.***>, 1 Eyl 2023 Cum, 04:48 tarihinde
şunu yazdı:

PR #488, opened 2023-09-04
Changed the way Paste keybind works in Composer

gaussandhisgun:
As it is, the Paste keybind in Composer's attachment page relies on `Gdk.Key.*` and `keyval`s. But [as it turns out](https://discourse.gnome.org/t/the-values-of-key-identifiers-in-gdk-depend-on-the-language-layout/7225), that makes alphabetical keys based keybinds not work at all on non-latin layouts (which i am a user of), or shift around on layouts that aren't default QWERTY. And while that [makes sense for, say, Dvorak](https://ux.stackexchange.com/questions/30666/keyboard-shortcuts-on-non-qwerty-keyboard-layouts), for Cyrillic it makes the users like me switch to a different keyboard layout just to paste an image. Which, again, is [against Gnome's principles](https://developer.gnome.org/hig/principles.html#reduce-user-effort). So, I did the only right thing that came to my mind, and added an `if` that also checks for `keycode` - so pressing Ctrl+cyrillic М or Ctrl+whatever-is-in-the-place-of-V will also paste the image. Keeping the keyval approach because of all those poor Dvorak people who think about what keys they press and not where they are located.

P. S. This is my first pull request. No, really, I've just figured out how to `git push` an hour ago. Probably would be a good idea to do something like this for other keybinds?


GeopJr:
Thanks for making this PR!

I don't really like using the physical key's location as a trigger. While it makes sense for QWERTY, other layout users (as you and the linked thread mentioned) don't mentally map shortcuts to their QWERTY locations but rather create new shortcuts. Additionally, there's no way to write docs for it (since it's position-based).

Having both the keyval and the keycode as triggers is not appropriate either. When their physical locations do not match (not QWERTY) there will be 2 different shortcuts that do the exact same thing, one of which might be already taken.

Would you mind if I postpone this for a bit so I investigate some more? I'm leaning towards making it an action and letting GTK deal with it or using the key translation functions (`Gdk.Display#translate_key`, `Gdk.Keymap.translate_keyboard_state`).

[off-topic]
1. Great work on your first PR!
2. You don't have to defend bug fixes 😅, that HIG section is about workflows, not... bugs

[/off-topic]

GeopJr:
Ended up using an action, thanks for the info and help!

PR #507, opened 2023-09-17
feat(preferences): organization

GeopJr:
I still think that there are too many options, especially under Appearance but it's somewhat more organized


LukaszH77:
I like it! 

If you think there are still too many options, maybe Appearance could get it's own tab?

GeopJr:
Ideally I'd like to split Appearance into smaller groups (on its own tab), I'm unsure of the groups:

pref

`#ff0000` fonts
`#00ff00` emojis
`#0000ff` media
`#ff00ff` posts?

but some of the rows don't really fit :/ letterbox media and expand pictures in media viewer are both about media but on very different contexts (posts vs media viewer). Fonts and emojis could also be included under posts, color scheme will be a single row

Looking at other apps, Megalodon groups them into three categories: Visual Appearance, Behavior and Timelines. I think that also suffers from this.

Mastodon web is all-over the place. Elk splits it into Interface (for colors, fonts) and Preferences (for everything else)

GeopJr:
I'll merge this as is at the moment and will revisit it again in the future!

PR #508, opened 2023-09-17
feat: don't include pre-installed icons

GeopJr:
Hopefully I won't regret this. Some icons are used with different meaning (repeat => reblog), have different color (non-starred) or their adw variants are different (warning), those stayed included.

TODO:
test on Ubuntu to see if their icons match


GeopJr:
I regret it but I'll move forward with it

Screenshot from 2023-09-17 21-44-54

PR #527, opened 2023-09-20
feat: DEVMODE

GeopJr:
Taking screenshots for the stores is awfully tedious, I usually run a Mastodon instance, create accounts, set them up, create the relationships, create the posts and quickly take screenshots. With DEVMODE, there's a special dialog that allows you to set some stuff up and push custom entities

image

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -42,6 +42,11 @@ if devel
   endif
 endif
 
+devmode = false
+if devmode
+  add_project_arguments(['--define=DEV_MODE'], language: 'vala')
+endif
+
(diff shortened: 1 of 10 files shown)


LukaszH77:
How does one enable this mode? Is it something to switch on while building the app, or an argument to add while starting it?

LukaszH77:
Nvm, found it :)

GeopJr:
WARNING:
I only made this to make taking screenshots easier, it disables saving settings and the websockets, probably shouldn't enable it outside of testing!!

PR #549, opened 2023-09-24
feat: add shortcut for Preferences

TheEvilSkeleton:
Per the [HIG](https://developer.gnome.org/hig/reference/keyboard.html), Ctrl+comma should open the Preferences.


GeopJr:
LGTM, thanks!

Issue #589, opened 2023-10-06
[Bug]: ctrl+enter in the post composer doesn't work when caps lock is on


nanopone:
### Describe the bug

ctrl+enter, the shortcut to send a post while in the post composer, fails when caps locks is toggled on. instead, it just makes a new line

### Steps To Reproduce

1. open the post composer
2. write something
3. toggle caps lock on
4. try to send the post with ctrl+enter

### Instance Backend

glitch-soc

### Operating System

fedora workstation 38

### Package

Flatpak

PR #590, opened 2023-10-06
fix(Composer): allow caps lock on ctrl + enter

GeopJr:
fix: #589


GeopJr:
If this causes issues again, let's make it an action

PR #593, opened 2023-10-08
Enhancement: Apply window maximized state from previous session to new session

henryrov:
Remembers whether or not the window was maximized when last closed.


GeopJr:
LGTM, thanks!

PR #602, opened 2023-10-12
feat(MediaViewer): ScaleRevealer

GeopJr:
Ported ScaleRevealer from Fractal

Very WIP as the way the media viewer is displayed changed. The style difference in the recording is not on purpose. Media Viewer API needs a refactor asap.

The animation looks really nice!

https://github.com/GeopJr/Tuba/assets/18014039/05b0e615-fbc1-433e-922e-3654e97b0549

TODO:
- ~~Refactor media viewer~~ kind-of
- ~~add comment~~
- ~~add other animations~~
- ~~make sure the mediaviewer headerbar follows the animation~~
- ~~swipe opacity of floating buttons and carousel indicator dots~~

### DIFF
--- a/data/gresource.xml
+++ b/data/gresource.xml
@@ -69,6 +69,7 @@
         <file>gtk/dropdown/expiration.ui</file>
 
         <file>ui/views/base.ui</file>
+        <file>ui/views/media_viewer.ui</file>
         <file>ui/views/profile_header.ui</file>
         <file>ui/views/sidebar/view.ui</file>
         <file>ui/views/sidebar/account.ui</file>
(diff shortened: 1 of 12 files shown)


LukaszH77:
I like it a lot, well done! However, I would prefer the background not to be semi transparent because the content visible behind the image distracts from the image and makes the controls less noticeable.

BTW, is there a chance for the media viewer to have two-clicks-to-zoom-in but one-click-to-close behavior, so I don't have to click that small < icon every time? :)

GeopJr:
https://github.com/GeopJr/Tuba/assets/18014039/47d2a4ca-b69e-4a33-bd89-10d31ad0a5c2

It should be close to done now, just some small paper cuts, let me know if you see anything wrong!

It's not only the animation, you should now also be able to dismiss it by dragging it vertically as shown in the video above!

Issue #136, opened 2023-03-29
[Bug]: Stream timeline doesn't refresh after the sleep of the computer


lapor-kris:
### Describe the bug

After sleep of the computer (Fedora 37), the stream timeline (home) doesn't get refreshed. As if the connection is lost.

### Steps To Reproduce

1. Open Tuba
2. put the computer to sleep (close the lid)
3.wake the computer (open the lid)
4. check the Tuba app (the stream is not refreshed)

### Logs and/or Screenshots

```
<paste your logs here>
```

### Instance Backend

Mastodon

### Operating System

Fedora 37

### Package

Flatpak

### Troubleshooting information

If I force quit the app through the System monitor in Gnome and open the app again, it works.

jdkruzr:
confirmed. also made me wonder why there is no "refresh" button!

nekohayo:
Possibly related: #204

ckruse:
@jdkruzr there is now refresh button, but you can hit ctrl+r to reload

primalmotion:
where is that refresh button? I don't see it on mobile

GeopJr:
`main` has pull to refresh, I'll make a new release soon enough after fixing some more issues!

primalmotion:
awesome! thanks!

primalmotion:
additional question, pull to refresh is nice to have, but is just a workaround to the connection loss after sleeping. I think it should eventually be fixed, but the question is: after pulling to refresh, will the push channel recover?

GeopJr:
Nope it won't. (Well, if it's part of a non-main timeline, you can go back and re-open it and it will re-establish a websocket connection)

This issue has some wider problems:

1. It's not that easy to reproduce, sometimes it reconnects, sometimes it doesn't disconnect at all (for me at least)
2. It might be a network connectivity issue. If a connection closes from the server, Tuba will keep retrying to connect every x seconds but that's as long as `websocket_connect_async` does not error. It could error for several reasons, one of them being lack of internet access (probably the cause of this issue) and the solution to this involves #204 (do we keep trying forever? do we create a queue and process it when there's internet connection again?)

I suppose a middle ground of checking if there's an active connection on refresh (and connecting) is probably a quick temp solution for this

primalmotion:
[...]

PR #616, opened 2023-10-23
feat: network cleanup

GeopJr:
libsoup can handle caching. Caching entities was more of a headache than helpful (any changes wouldn't reflect or would reflect globally). Caching images in memory for x amount of time is wasteful.

Following tootle discussions it seems that it was implemented because libsoup's wasn't reliable. Tuba migrated to libsoup3 however which might have fixed it.

TODO:
- ~~Check if cache actually works reliably~~
- Cleanup
- ~~Remove abstract cache altogether~~
- ~~Cache blurhashes?~~
- Check reactive entities (like pinning statuses)
- ~~clear cache on close?~~

fix: #136


GeopJr:
Libsoup does in fact hit the cache, or at least the logger claims so.

I simplified the API by getting rid of "is_loaded" and instead kept track of it locally. It also takes care of blurhash too. If a blurhash is provided, it will idly decode it and return it if the image hasn't loaded yet.

GeopJr:
Memory leak in libsoup?
is this expected?
what's going on?

image

GeopJr:
I think that last commit fixed #136 :shrug: needs more testing

GeopJr:
... memory leak with FlowBox#bind_model :shrug: used helper functions instead

GeopJr:
I'll go ahead and merge it. There's no way to find out what might be wrong unless it starts getting used outside of this PR

LukaszH77:
I used it for some time and it worked fine.

GeopJr:
It has some problems here and there, especially:
- memory leaks became more prominent
- sometimes it just gives up and doesn't recover - I can reproduce it sometimes on the custom emoji picker, where it will stop midway and then no other images will load at all

Overall this change is actually positive despite its shortcomings. From a user standpoint, it saves on bandwidth by caching images and metadata locally (`.cache/Tuba/soup/...`) and doesn't clog the main connections by using a separate session for images. From a dev standpoint, it finally gets rid of the whole entity cache headache and should allow me to move forward with other plans.

Thanks for testing it as always 💛🎺

PR #617, opened 2023-10-24
Use copy icon for copying handle

kra-mo:
The share icon is more commonly associated with opening a share sheet and with no visual feedback of a successful copy, it can look like the button did nothing.

Ideally, the copy action would display a toast too.


GeopJr:
LGTM, thanks!
I'll open another issue for feedback toasts

Issue #612, opened 2023-10-18
[Request]: Compose button on the conversation view


Nico7as:
### Describe the request

Hi 
Please note that I use the french version of Tuba, so I might refer to some functions or buttons with an aproximative name.

Please add a "compose/new message" button on the conversation view.
and then make the new message "direct" or "private" by default.

I would be very convinient to use Tuba for direct messaging.

### Implementation Details

- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

henryrov:
I think this feature makes sense, and it's simple enough to implement. I opened #619 as a draft implementation, if anyone wants to go over it.

GeopJr:
I'm not against this but also not really sure if it's that useful. Conversations are (as already mentioned in the OP) just another visibility option so the current open view doesn't really matter where, for example, on platforms like Twitter, the "Direct Messages" view is its own thing, that can only be accessed there and can only be used for chatting. On fedi instead, you can have 'direct' visibility under any post, including public ones without the author of it being able to even see the message. Those also end up in the Conversations view.

What I'm trying to get at is that the question is not about whether the conversation view should have the button but whether *all* (or at least most) views should have it.

FWIW, our designers were clear that the floating button is only meant for the "Home" view. (That is not set in stone of course).

- Elk.zone seems to also follow that where the composer is only visible on "Home"
- Mastodon web has an always visible sidebar composer on desktop and a headerbar(?) button on non-desktop, visible on all views
- Megalodon doesn't have a conversations view but the floating button is visible in multiple places (the main view, hashtags, profiles... but NOT in threads, notifications...)

Also quick reminder, the original issue is only about *creating new conversations*, the button would not be used for replying. Click the reply button under the conversation you are replying to instead, if that's what you are trying to do.

nekohayo:
> […] the question is not about whether the conversation view should have the button but whether all (or at least most) views should have it.
> 
> FWIW, our designers were clear that the floating button is only meant for the "Home" view. (That is not set in stone of course).

In case this is useful, I will say that in my personal experience, I have often (every couple of days) found myself frustrated by not being able to compose a new independent toot without exiting the current view (thread, notifications, etc.). It is annoying because it is artificially limiting me into a single-tasking mode. I am then:
* Unable to keep the context of what I was looking at that inspired me to post something new
[...]

Issue #548, opened 2023-09-23
[Request]: Compose button: reduce animation wonkiness, improve discoverability


bragefuglseth:
### Describe the request

The compose button is currently hidden when the window is scrolled downwards. I can see why this is done; it makes the view more focused. However, the button currently just slides into nothingness. In addition, the Go to top button doesn't animate, it just jumps. Both of these things look weird, especially when combined:

[Screencast from 2023-09-23 15-38-00.webm](https://github.com/GeopJr/Tuba/assets/91388039/8b1ca969-bc92-456c-9883-e045aaab1203)

As an initial solution, I think this would work better if:

1. The compose button disappeared into the bottom of the window instead of having its animation cut off
2. The Go to top button also animated instead of just jumping

I'm a little concerned about the compose button only being present when the view is scrolled upwards in general, though, as it's not obvious, and will lead to people questioning where it is. I think we should consider always having it visible.

To reduce clutter without adding confusion, what about toggling the visibility of the Go to top button instead? We can hide it by default and only display it when the view is scrolled upwards, since that implies that people might want to go all the way to the top.

### Implementation Details

- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

GeopJr:
> I'm a little concerned about the compose button only being present when the view is scrolled upwards in general, though, as it's not obvious, and will lead to people questioning where it is. I think we should consider always having it visible.

The behavior is supposed to mimic the android one (scroll down => hide instantly, scroll up => show after some distance). Perhaps we should only have this behavior on narrow mode (?) and have it always visible on wide? I'm ok with either!

> To reduce clutter without adding confusion, what about toggling the visibility of the Go to top button instead? We can hide it by default and only display it when the view is scrolled upwards, since that implies that people might want to go all the way to the top.

FWIW, the scroll to top button is far less noticeable on views that don't have the floating compose button (/ not Home)

![image](https://github.com/GeopJr/Tuba/assets/18014039/d61bfcd3-ad6f-4163-8a30-e272e6788b93)

bragefuglseth:
> Perhaps we should only have this behavior on narrow mode (?) and have it always visible on wide?

Sounds good to me, as long as the animations are fixed 😁

GeopJr:
Does it look right?

[Screencast from 2023-09-25 01-27-59.webm](https://github.com/GeopJr/Tuba/assets/18014039/705cab55-c827-4cc7-a907-e35696b492de)

[...]

PR #619, opened 2023-10-24
Compose button in direct messages

henryrov:
This is a draft implementation of the feature requested by #612. Some considerations:
- I'm not sure if adding another action is the most desirable way to achieve this, but it plays nicely with the existing codebase
- What button icon to use
- Not specifically related to this issue, but maybe add a scroll-to-top button to the messages tab in the future
Of course, if this feature isn't thought to be necessary or make sense within Tuba at the moment, feel free to close this.

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -346,7 +346,7 @@ namespace Tuba {
 			new Dialogs.Compose ();
 		}
 
-		void compose_direct_activated() {
+		void compose_direct_activated () {
 			if (accounts.active.instance_info == null) return;
 
 			new Dialogs.Compose.direct ();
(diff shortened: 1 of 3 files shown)


GeopJr:
Thanks for this PR!

From a quick look the main concern is the floating button. It's a bit more complex than that:
- It has to play nice with the scroll to top button (on Home, the scroll to top button moves above it)
- On narrow mode (/ mobile), it has to disappear when scrolling down and re-appear when scrolling up, but not instantly, you have to scroll up a bit more (following android's behavior)
- On narrow mode it has to be visible when near the top
- It needs to be inside a revealer so there's an animation when it (dis)appears
- There needs to be an animation when the scroll to top button is also visible (https://github.com/GeopJr/Tuba/issues/548#issuecomment-1732684520)

Depending on how the discussion on #612 goes, ideally, the code from the Home view should move either to the Timeline view (visible in all timelines) or to a new FloatingButtonView so both Home and Conversations can inherit from it instead of having to duplicate the same complex logic for each one

Issue #618, opened 2023-10-24
[Request]: feedback toasts


GeopJr:
### Describe the request

Show toasts when appropriate for feedback (e.g. when a button does no visual changes)

from: #617 

### Implementation Details

- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

PR #620, opened 2023-10-24
feat: global toast feedback

GeopJr:
fix: #618 

not sure where else to use them:
- Copied handle
- Copied url
- Copied media
- Saved media
- Errors (no timeout)


GeopJr:
We can think more later

Issue #553, opened 2023-09-25
Composer redesign


GeopJr:
The composer is a bit complex. We can divide it into 3 sections: Text, Media and Polls

- Each section has its own set of actions
- There can be either Media or Polls
- Media can be published on their own but polls require some Text
- Text character counter includes the Content Warning
- Text character counter replaces urls with a set amount of characters (not done in tuba yet, depends on instance's info, ex. `https://geopjr.dev` is 18 characters but will be counted as 7, the regex should match mastodon's so we match them 1:1 even if it has its quirks)

Below I've attached screenshots from other apps' composers. It's important to notice how the pages and actions interact with each other. Megalodon and Mastodon web make the inactive actions insensitive (the media attach button is no longer sensitive when polls are active), Elk replaces them (the media attach button gets replaced by the poll's actions) and tuba has a pages system

|App|Text|Media|Polls|
|:---:|:---:|:---:|:---:|
|Tuba| ![Screenshot from 2023-09-25 13-00-08](https://github.com/GeopJr/Tuba/assets/18014039/5997cc63-ecff-4747-bb0c-1503138e6908) | ![Screenshot from 2023-09-25 13-00-18](https://github.com/GeopJr/Tuba/assets/18014039/e698c21b-b337-427b-97c1-a0ad51a63f64) | ![Screenshot from 2023-09-25 13-00-25](https://github.com/GeopJr/Tuba/assets/18014039/518b8a13-c7a8-48e2-8ab1-6a08823ee1bf) |
| Megalodon | ![Screenshot_20230925-125433_Megalodon](https://github.com/GeopJr/Tuba/assets/18014039/768fe2fc-5ac0-4d34-98de-c840bb4e7f63) | ![Screenshot_20230925-125505_Megalodon](https://github.com/GeopJr/Tuba/assets/18014039/ee172db9-494e-46f7-9ad5-bcb3422ab588) | ![Screenshot_20230925-125518_Megalodon](https://github.com/GeopJr/Tuba/assets/18014039/e5b79e34-e706-475c-871b-e00fd3ff2308) |
[...]

Issue #472, opened 2023-09-01
[Request]: Enhance the media support


gaussandhisgun:
### Describe the request

So as I wrote in #468, pasting images is a nice feature. #413 (As It Obviously Turns Out) adds that, so I installed nightly flatpak to try it out. And IMHO, pressing Ctrl+V on the "Text" tab when an image is in the clipboard should go to the Image tab and do the paste there. Telegram for Android behaves *exactly* like this if you paste an image to it. Mastodon and Firefish (both of backends I've tried) do not require you to open the embed menu to paste an image either. Reddit, no matter how bad it is, just puts the image *there* if you Ctrl-V an image into the Text post. You name it.

### Implementation Details

- [ ] This should be an option in settings.
- [ ] This should be only available to some fediverse backends. (Include which ones on the above field).
- [X] This is client-only (and shouldn't sync with the instance).
- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

GeopJr:
I'm not super-fond of this to be honest due to the way media-uploading works.

When you attach a file, it first gets uploaded to the server which returns an id that we later attach to the object we send when "Post" is clicked. I wasn't super-fond of Ctrl+V for the same reason but at least the user is somewhat aware by being on the Media page of the composer.

That means that if you accidentally press ctrl + v with sensitive media in your clipboard, they will be uploaded to your instance which has privacy issues. At the same time however, you are supposed to trust your instance somewhat and they do get cleaned-up after some time automatically.

At the same time, when the new composer mockups drop, there probably won't be a "pages" system anyway. The solutions I propose are:

- A dialog confirming the paste on Ctrl + V
- Assume instances and their sysadmins do not lookup or keep the discarded media forever

What do you think?

LukaszH77:
I like the option with confirmation better, but I think for some people that could be annoying

Maybe you can add an option to switch it off?

bragefuglseth:
> Maybe you can add an option to switch it off?

Obligatory link to [Choosing our Preferences] (https://ometer.com/preferences.html)

jmaris:
I do worry that creating a preference for this would be overkill. I'm not sure it's guideline compliant, but perhaps a confirmation dialogue with the yes option focussed by default. That way people can CTRL+V then ENTER to proceed.

Baralheia:
[...]

PR #623, opened 2023-10-28
feat: new composer

GeopJr:
fix: #553 
fix: #973
fix: #472
fix: #1453


GeopJr:
Judging by the errors, we are going to have to bump some dependencies, I'll make a last patch release when the current release is ready for the stable distros but that's it. I'm not going to hold back the new composer any longer and all the stuff used are **needed**.

bertob:
I tried the branch, mostly looks great to me! Thanks so much for working on this :raised_hands: 

A few minor issues I came across:

The alt text indicator can be hard to read depending on the image, perhaps we could make it more (or fully?) opaque? It also becomes more transparent on hover, which looks odd.

image

image


image

The alt text below the image in the editor looks a bit lost layout-wise, since it's not anchored to the content. 

Screenshot From 2025-06-22 12-02-23

At small sizes it's also a bit odd that the alt text scrolls on its own. I think it'd be better to scroll the entire view, including the image and the row below with the character count.

image

It's not ideal that the single choice / show results labels are ellipsized. I realize now that this was a flaw in the mockups, since I never actually tried drawing this for phone sizes. Maybe we could just move the third item to a second row and make it full-width though?

Screenshot From 2025-06-22 12-22-19

When opening the scheduling dialog it has a different width than the main post dialog:

[Screencast From 2025-06-22 12-47-39.webm](https://github.com/user-attachments/assets/89c754eb-a6a4-4a10-9e7e-ee14b8715bbc)

GeopJr:
Thanks for the review :pray: 

> The alt text indicator can be hard to read depending on the image, perhaps we could make it more (or fully?) opaque? It also becomes more transparent on hover, which looks odd.


| | |
|---|---|
| Screenshot From 2025-06-22 14-01-36 | Screenshot From 2025-06-22 14-01-26 |
| Screenshot From 2025-06-22 14-38-15 | Screenshot From 2025-06-22 14-37-57 |

I think the second row is better IMO, more opaque + blended with white/black while keeping the text color. (FWIW, in high contrast the indicator becomes completely opaque etc.)

> It also becomes more transparent on hover, which looks odd.

My reasoning was that since it's not a button, having it become more transparent would better show that it's just an indicator and by hovering the user can see the attachment below it better. Not strong opinion on it, I'll remove it!

> The alt text below the image in the editor looks a bit lost layout-wise, since it's not anchored to the content.

Can you expand on what you mean by 'anchored to the content'? / How it can be fixed?

> At small sizes it's also a bit odd that the alt text scrolls on its own. I think it'd be better to scroll the entire view, including the image and the row below with the character count.

I agree on the character counter but I'm not so sure about scrolling with the attachment. I feel that if you are writing alt text for an image, you'd want to keep looking at it instead of scrolling up and down to see and describe it. Mastodon mobile seems to do that too, the image can be scrolled but the alt text box stays in place at the bottom.
(The character counter could also be moved to the headerbar or inside the image as an overlay)

> It's not ideal that the single choice / show results labels are ellipsized. I realize now that this was a flaw in the mockups, since I never actually tried drawing this for phone sizes. Maybe we could just move the third item to a second row and make it full-width though?

||||
|---|---|---|
| Screenshot From 2025-06-22 14-23-56 | Screenshot From 2025-06-22 14-23-32 | Screenshot From 2025-06-22 14-22-42 |

I think it looks a bit awkward on the non-bottom-sheet versions. Maybe keep it as 3 columns there? It would ellipsize though... Alternatively I can increase the default window width.

GeopJr:
- Attachment editor

Added a light shade background + paddings and added a clamp for the alt text. Not sure if it's a good idea however and might be better to do only horizontal margins. Counter moved to the headerbar. FWIW, videos and audio show a video player instead of a focus picker.

|||
|---|---|
| Screenshot From 2025-06-24 04-15-06 | Screenshot From 2025-06-24 04-15-00 |

Screenshot From 2025-06-24 04-15-48


- Polls

|||
|---|---|
| Screenshot From 2025-06-24 04-25-22 | Screenshot From 2025-06-24 04-25-02 |
| Screenshot From 2025-06-24 04-24-16 | Screenshot From 2025-06-24 04-29-43 |
| Screenshot From 2025-06-24 04-49-46 | Screenshot From 2025-06-24 04-49-19 |

The first three are no-arrow with different alignments, the fourth one is no-arrow and short label and the other two are icon-only show results with different alignments.

I kind-of see only the last two being the least awkward... I don't think 3 long-ish labels will fit nicely in that space and it might be even more jarring with other languages that tend to have longer words :/

---

Pushed the better colored indicators (and implemented the validity checks so the 'success' variant can be tested now). I did notice **two things I missed**:

1. The old composer has a button for marking media as sensitive which the mockups don't. Should I add it to the bottom bar? header bar? below the attachments?
2. the mockups have a unique symbolic for the polls which is not present in library, if you have it at hand, please send it!

bertob:
### Image editor

> Added a light shade background + paddings and added a clamp for the alt text. Not sure if it's a good idea however and might be better to do only horizontal margins. Counter moved to the headerbar. FWIW, videos and audio show a video player instead of a focus picker.

It looks a bit odd that the text is not aligned with the image, maybe from that PoV it'd be better to just do full-width? The horizontal padding is also a bit too wide now, I'd go with a similar horizontal and vertical padding on the textarea (e.g. 12px).

I'd also increase the vertical padding around the image to 12, it looks a bit cramped currently.

As for the counter, on the top left it looks like it belongs to the back button, which is not ideal. Perhaps we could drop the view title and move it to the right? Could also try dropping the bold and maybe making it dimmed, to make it clearer that it's not a button. I'd also add a tooltip like `Remaining Characters`.

image

More generally, it's a bit of an odd pattern to have a sub-view with a back button that deletes all changes unless you hit "save". Is there a reason why we can't just insta-apply the changes without the need for a save button?

### Polls

I'd try going with this for the narrow case, or maybe this but with the date thing using half the width and left-aligned:

image

and this (with the arrow) for the wide case - maybe with an even narrower duration picker, it's fine if that one is ellipsized a bit in some cases since it'd just be the last few characters of "minutes":

image

---

> The old composer has a button for marking media as sensitive which the mockups don't. Should I add it to the bottom bar? header bar? below the attachments?

In the old composer it looks like it's per-post rather than per-image, is that how the setting works? Otherwise maybe it could be an overlay with an eye icon in the top left of the thumbnails?

> the mockups have a unique symbolic for the polls which is not present in library

polls-symbolic

bertob:
### Image Editor
The latest version looks good to me, though I would add back a title (maybe just "Edit Media") since we have enough space now.

### Saving to server
More of a general point and I'm not sure how actionable it is, but I'm really not sure about sending anything to the server until you actually submit the post (I just realized now that images are actually immediately uploaded when you add them). From a user point of view this seems counterintuitive in a native app (rather than a website, where the assumption is that anything you do goes to the server).

No idea what our options are here, but maybe worth considering saving drafts locally only until you actually send/explicitly save as draft on the server?

### Polls
That video doesn't work for me in multiple (Flatpak, so not a codec thing I think?) browsers and downloading doesn't work either :(

### CW for media
Can we just add another icon button (maybe with the strikethrough eye) next to the other CW button then?

GeopJr:
# Saving to server

It could be possible for the alt text + focus but the media is a sticky one. When we send the final post to the server we attach the uploaded media ids, which means we first have to upload them separately. The issue is that uploading can fail for a number of reasons outside of our control (we already check for size and what's allowed etc). For example Mastodon often drops (dropped?) videos that had weird framerates. So the user would have to deal with these when pressing 'Post'. Other than that, there's some strict rate limits around media, so uploading them all at once can reach them fast (often what happens on the tuba@floss.social announcements and the threads stay unfinished until the limits pass).

I'll do a bit more research on this but I'm not too positive it will work reliably.

# Polls

ScreencastFrom2025-06-2705-48-59-ezgif com-video-to-gif-converter

# CW for media

It should only be visible when there are attachments, right?

bertob:
I tried the grid options, none of them are great but the left aligned one is the least bad one for the narrow layout. I think we could go with that for mobile, but then switch to the three column layout (with a narrow third column) as soon as possible. Maybe we could also try to switch to shorter strings in the three column case, e.g. using "min" instead of "Minutes" and "h" instead of "Hours", to minimize the width of the third column.

image

> It should only be visible when there are attachments, right?

Yup, that's what I was thinking.

bertob:
The poll thing looks good to me layout wise, but I'd go with more standard abbreviations, i.e. lowercase letter for hours and days, and not abbreviate minutes to "M". So e.g. `5min`, `6h`, `3d`.

The sensitive media setting looks good, but it'd be nicer if it actually previewed that the images will be blurred. Could we already blur the thumbnails in the composer when the setting is on to make it more obvious what it does?

bertob:
LGTM :)

PR #626, opened 2023-10-31
fix(LWW): do not ellipsize when lines && widgets > 100

GeopJr:
Problem: without setting lines to 100 and ellipsizing, LLW would mess up its sizing. I do not think this is a problem anymore, maybe, but without ellipsizing theres a performance impact so it has to stay for now until the switch to ListView or other optimized containers. However, it seems that when a post contains only custom emojis, it needs ellipsizing to be off for the wrapping to be done right (ex. https://eepy.moe/notes/9li57c0uugndgjq2)

Solution: when lines and widgets > 100, set ellipsizing to none :shrug:


GeopJr:
Closed in favor of #644 for now

Issue #423, opened 2023-08-01
[Request]: (Un)Follow / "Follows you" and "In X lists" actions and indicators in lists of users


nekohayo:
### Describe the request

I'm not sure how exactly this would look in practice in the UI, but I'd like the ability to look at other people's list of followers (or their lists of users they follow) and to know which ones I already follow or not, which ones follow me, what lists of mine they might be in (or at least how many?), etc. so that I can follow people and  add them to the correct lists when I don't already follow them.

Currently, when I go to someone's profile and visit the "Following" and "Followers" sections of their profile, I see a list of users, but the only possible actions in Tuba 0.4.0 are "Open in browser" and "Copy URL", hidden in the meatballs menu for each user row.

### Implementation Details

- [ ] This should be an option in settings.
- [ ] This should be only available to some fediverse backends. (Include which ones on the above field).
- [ ] This is client-only (and shouldn't sync with the instance).
- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

GeopJr:
Sounds good! We need to make another request with all user ids to get our relationships (whether we follow them, they follow us, we are blocked, they are blocked etc) with them. I'm not sure about lists however, I don't think there's a way to bulk check if users are in any and keeping all lists and their members in memory seems wasteful (especially for accounts with many lists and members).

In the meantime, clicking on an account row in the following/followers lists, should open their profile where you can follow/unfollow from and add/remove them to/from lists

GeopJr:
Update:
#637 

The current design is:
![Screenshot from 2023-11-11 04-32-31](https://github.com/GeopJr/Tuba/assets/18014039/3c585fc1-1806-4f79-9714-08141b5fa268)

As mentioned in the PR, not sure if there should be a compact, ActionRow-like alternative layout since that might be a bit too much information for the following/followers lists :shrug: 

About the "In X lists", the previous comment still applies - there's no way to bulk retrieve that info. However, we could maybe keep in memory the list info like { list_name => [account_ids] } and check manually but would definitely need more thought on how to approach it and edge cases

PR #637, opened 2023-11-11
feat: new account widgets

GeopJr:
fix: #423

- Inspired by elk's
- Not sure if there should be a smaller alternative one that resembles Adw ActionRow
- Added bulk relationship resolving
- Cleanup some APIs
- Depends on #616 due to the entity cache

TODO:
- ~~Use in notifications~~
- ~~Cleanup the timeline api~~
- ~~Check if it's leaking~~

Screenshot from 2023-11-11 04-32-46
Screenshot from 2023-11-11 04-32-31


GeopJr:
im going to ship this even if it's last second

PR #644, opened 2023-11-17
feat(LWW): remove hack

GeopJr:


bugaevc:
If you're no longer setting `label.ellipsize` to anything but `NONE`, you should be able to remove more of this code (`old_ellipsize`, `new_ellipsize`, ...)

GeopJr:
Done!

Issue #604, opened 2023-10-16
Back button on top-level views


bertob:
In wide mode all top-level views except "Home" have a back button, which is a bit odd because they're at the same level of hierarchy as Home. I'd just drop them.

bertob:
The Search view still has the back button for me on the latest main:

![image](https://github.com/GeopJr/Tuba/assets/1908896/b3a72fa8-96b9-42fe-9153-7a1fb5f1adfe)

GeopJr:
FWIW, the search view can be activated with <kbd>Ctrl</kbd><kbd>F</kbd> from any other view, that one *should* have the back button, right?

GeopJr:
Between not all servers having post search enabled and users having to opt-in to get theirs indexed, I'm leaning towards removing <kbd>Ctrl</kbd><kbd>F</kbd> altogether.
Otherwise we could search based on what Tuba has in memory

bertob:
If it worked I think a per-view search feature would be cool, FWIW. I'm constantly looking for old posts of mine and can never find them :(

PR #648, opened 2023-11-19
feat: remove back arrow from sidebar items

GeopJr:
including the fake ones

fix: #604


danialbehzadi:
If back buttons are removed, there should be sidebar button or user will be stuck on that page for ever!

GeopJr:
The back buttons are removed from *sidebar* items only, these ones:
image

The reason is that there can only be one active at a time anyway. If you are at "Home" and select "Bookmarks", you can go back by selecting "Home".

Nested pages still have back buttons, e.g. "Home" => "Bookmarks" => open a post => "Thread" (has back button) ===back===> "Bookmarks"

GeopJr:
I see what you mean now (/ that it only happens on mobile/narrow mode), thanks for letting me know!

Issue #660, opened 2023-11-24
[Bug]: Hashtags autocompletion / keyword popularity statistics popover can sometimes break


nekohayo:
### Describe the bug

Sometimes, the autocompletion widget for hashtags stops working.

This bug has been annoying me for months, but I had never observed exactly what causes it to happen, so I recorded a video below to be able to see when it does and doesn't trigger.

### Steps To Reproduce

There are a few ways, including:
* Replacing characters (with the delete key in the middle of a string, or maybe overwrite insertion mode)
* Selecting (with mouse or keyboard, such as ctrl+shift+left) a portion or entirety of the hashtag and trying to overwrite it. See video below.
* Edit: also by adding the `#` hashtag symbol after having written the hashtag's word string

### Logs and/or Screenshots

In this video, pay attention to these times in particular:
* 1 minute 46 seconds to ~1 minute 55 secs
* 2 minute 50 seconds to the end

https://github.com/GeopJr/Tuba/assets/479401/73f9daea-d682-4473-840d-23ef46b4bda4

### Instance Backend

Mastodon

### Operating System

Fedora 39

### Package

Flatpak

### Troubleshooting information

os: GNOME 45 (Flatpak runtime)
prefix: /app
flatpak: true
version: 0.5.0 (production)
gtk: 4.12.3 (4.12.1)
libadwaita: 1.4.0 (1.4.0)
libsoup: 3.4.3 (3.4.3)
libgtksourceview: 5.10.0 (5.10.0)

### Additional Context

I'm wondering if this might be related in any way to #207 or the refactoring that would be part of #153 …

nekohayo:
Oooh, I found another way I _very_ frequently break hashtags autocompletion: by writing the word first, and then adding the `#` symbol in front of the word (because I often remember to add hashtags after I wrote the post)... doing so will break that hashtag's interaction until the end of the player/character's turn ;)

GeopJr:
#661 should probably fix both, even though I'm not sure I understood the first one completely. 

The strategy prior to that PR was (well, gtksourceview kind-of pushes you to do this with the way get_word works) to get the start of the word *until* the cursor. It now gets the actual whole word. Other things I noticed that bugged me:

1. Due to #207 it would return no results at first if it found the trigger (probably what you are experiencing in the screen recording) and only show them after you modify the word again
2. If you modified a word it would replace part of it instead of the whole word (`#maaaastodon` == remove the `aaa` and select first result ==> `#mastodon aaastodon`)

PR #661, opened 2023-11-24
fix(CompletionProvider): refactor word capturing logic

GeopJr:
fix: #660 

among other things it also fixed the word replacement strategy by replacing the whole word instead of appending


GeopJr:
Found some more edge cases with replacing the word when there's suffixes and with word capturing when the last character is already the end of the word.

I couldn't find anything else

PR #678, opened 2023-12-01
feat(flatpak): remove webp-pixbuf-loader

GeopJr:
https://floss.social/@sonny/111505420648908207

We won't have to bundle it in the future, it might not be available on 45beta (the runtime the CI uses) but I'll leave this here as a draft


GeopJr:
Merged in #804

Issue #582, opened 2023-10-02
[Request]: make external link confirmation optional


teemuxyz:
### Describe the request

Tuba asks for confirmation every time user clicks on a link that opens in an external application. This adds an unnecessary extra step to the process. There should be a setting to turn this confirmation off.

### Implementation Details

- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

kra-mo:
From a design standpoint, this is one of those dialogs you start to instinctively click through after a couple of times. I definitely support adding an option to disable it.

PR #679, opened 2023-12-02
feat(PreviewCard): remove dialog

GeopJr:
fix: #582 
close: #666 (for now)

Upon further discussion, removing it altogether seems like the best approach


GeopJr:
reverted the string removal as it might come back in the future with different UX and shouldnt create more work to translators

Issue #665, opened 2023-11-26
[Bug]: In German, screen is cut off when window has small width


tronta:
### Describe the bug

When set to German language, the screen gets cut off in small window sizes/mobile:

![Bildschirmfoto vom 2023-11-26 08-04-35](https://github.com/GeopJr/Tuba/assets/5135577/612a983e-04e3-4397-be29-d74730c9e0d0)

[Bildschirmaufzeichnung vom 2023-11-26, 18-08-58.webm](https://github.com/GeopJr/Tuba/assets/5135577/b6a4621c-ae83-4474-b295-dc04f71c7784)

### Steps To Reproduce

1. Switch to Geman
2. open on mobile, or resize screen to minimum

### Logs and/or Screenshots

```
<paste your logs here>
```

### Instance Backend

Mastodon

### Operating System

Fedrora Silverblue 39

### Package

Flatpak

### Troubleshooting information

os: GNOME 45 (Flatpak runtime)
prefix: /app
flatpak: true
version: 0.5.0 (production)
gtk: 4.12.3 (4.12.1)
libadwaita: 1.4.0 (1.4.0)
libsoup: 3.4.3 (3.4.3)
libgtksourceview: 5.10.0 (5.10.0)

camelCaseNick:
The label “%lld voted” was literally translated to “%lld haben abgestimmt”. I have changed it to “%lld Stimmen" (literally "%lld votes”). Soon, Weblate should push my change. Additionally, I think #695 would be great to avoid this kind of issues in the future, because an ellipsized and thus truncated string is far less disruptive than a complete UI shift making parts of it unreachable.

PR #695, opened 2023-12-10
fix(votebox): ellipsize vote count and expiration label

camelCaseNick:
These labels can in theory become arbitrarily long due to translation. Thus, its width should be controlled using ellipses. This is as of now an issue with the German translation.

Fixes: https://github.com/GeopJr/Tuba/issues/665

### DIFF
--- a/data/ui/widgets/votebox.ui
+++ b/data/ui/widgets/votebox.ui
@@ -23,6 +23,7 @@
         <child>
           <object class="GtkButton" id="button_vote">
             <property name="visible">0</property>
+            <property name="valign">center</property>
           </object>
         </child>
         <child>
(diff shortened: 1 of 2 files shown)


GeopJr:
Thanks for taking a look into it!

I'm not super fond of ellipsizing it :/ Not sure how it would end up in German, but I'd find it pretty weird if it ended up as "1500 vo..." and "expired on 11..."

I'm more in favor of combining the two labels into one and enabling wrapping

GeopJr:
Pushed it, let me know what you think of it before merging!

camelCaseNick:
With “Middle Dot”, I think, it also looks better than the current version regardless of the wrapping / width issues. Then, another idea would be to wrap this label below the vote button before wrapping itself, but I think it works quite well this way.

> Not sure how it would end up in German

With the upcoming strings, it should roughly be:

grafik

This is definitely better than having ellipses.

GeopJr:
> Then, another idea would be to wrap this label below the vote button before wrapping itself, but I think it works quite well this way.

That would probably require a flowbox (like the name+handle one) which might have other issues, better revisit it after the next release

PR #701, opened 2023-12-13
feat(SecretAccountService): do not read all secrets

### LINKED ISSUE #114: [Bug]: it ask for server URL everytime
### Describe the bug

I am not on a Gnome based DE, so maybe a dependency is still missing. But I can't find something relevant in the logs.

Alpine recipe: https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/45396

It logs

```
Tuba-Message: 08:40:45.704: SecretAccountStore.vala:111: Saved secret for @stacyharper@soc.punktrash.club
Tuba-Message: 08:40:45.704: SecretAccountStore.vala:67: Saved 1 accounts
```

which seems to be a OK

### Steps To Reproduce

run tuba

### Logs and/or Screenshots

Two log files for two successive tuba process:

https://paste.sr.ht/~stacyharper/212868a852a2cd6f70f3b6efa5c12f833dc1adf5

### Instance Backend

Pleroma

### Operating System

Alpine Linux

### Package

aports

?

### Additional Context

_No response_

GeopJr:
To fix #114 we decided to do #134

Upon further usage of libsecret and further understanding of its internals, that's a terrible idea. Instead of just searching for the dev.geopjr.Tuba secrets, it looks up ALL of them that match the other attributes (version=1,login=\*). That can lead to other non-tuba secrets getting both read and parsed by tuba.

Ultimately, we need to revisit #114 and figure out why the `xdg:schema` did not match tuba's.

### DIFF
--- a/src/Services/Accounts/SecretAccountStore.vala
+++ b/src/Services/Accounts/SecretAccountStore.vala
@@ -13,7 +13,7 @@ public class Tuba.SecretAccountStore : AccountStore {
 		schema_attributes["version"] = Secret.SchemaAttributeType.STRING;
 		schema = new Secret.Schema.newv (
 			Build.DOMAIN,
-			Secret.SchemaFlags.DONT_MATCH_NAME,
+			Secret.SchemaFlags.NONE,
 			schema_attributes
 		);
 


GeopJr:
I can't reproduce #114 but it also didn't save the first time it ran, might be due to Kwallet asking me for the encryption method & passwords. It saved just fine on the second time though!

I tried it using KDE as #114 was about non-GNOME.

@stacyharper (if you are still using non-GNOME) whenever you get some free time, could you remove any tuba entries from your keyring and give this another try? just to make sure I don't accidentally break it in the upcoming release

stacyharper:
> @stacyharper (if you are still using non-GNOME) whenever you get some free time, could you remove any tuba entries from your keyring and give this another try? just to make sure I don't accidentally break it in the upcoming release

Yes it seems it works as expected. From locked and empty keyring, it connect correctly and store the entry. It can quit and restart Tuba, and it is still connected, even if I lock the keyring back in-between.

stacyharper:
@GeopJr I feels much sorry. But now Tuba ask for the server domain after every computer reboot. If I unlock manually the "Default Keyring" with Seahorse, it works. Even if I lock it afterward, Tuba continues to works, and open the prompter to type my passphrase. But on a fresh boot, after a computer reboot, it does not.

GeopJr:
We need to figure out *why* this is happening and *why* does it work if we pass DONT_MATCH_NAME. Can I have some more info so I recreate your environment 1:1?

Specifically:
- Distro
- DE
- Secret service / keyring
- Is the default keyring password the same as the user one?
- DM
- Any other info you think might be important

I definitely don't want to go back to `DONT_MATCH_NAME`. I'm working on another app that has the same attributes as tuba and `DONT_MATCH_NAME` ends up reading both of their secrets.
(Just to cross this out, I don't think it's a musl, alpine, non-gnu, non-systemd issue as my main machine runs chimera linux)

stacyharper:
I am using Sxmo, so Sway as WM. It is the same password than my user one, but we use tinydm to automatically start the session. It means there is no pam rules that trigger to auto-open the keyring. We expect a prompter when the first program as for a keyring entry. I let the keyring to be created automatically, meaning it is correctly named "Default Entry". Other programs correctly trigger the prompter.

(I'll try some refresh, so I'll edit this post)

afk-mario:
Same here using 

Distro: Arch
DE: Sway
Secret service: gnome-keyring

GeopJr:
Okay so here's what I've gathered:

The issue is https://gitlab.gnome.org/GNOME/libsecret/-/issues/7 and https://gitlab.gnome.org/GNOME/gnome-keyring/-/issues/89

Does this fix it? https://gitlab.gnome.org/GNOME/libsecret/-/merge_requests/105
Probably not (0.21.0 is on >= 3.19 on Alpine)

All the linked issues list as a workaround `DONT_MATCH_NAME`, but that is definitely undesirable, at least for Tuba (for the reasons mentioned above). A workaround for Tuba specifically would be to bump the schema to include something like `app=dev.geopjr.Tuba` BUT bumping it will probably cause all previous Tuba secrets to not match and require all users to re-login :shrug: 

My hands are tied, any suggestions?

GeopJr:
Maybe do a lookup with `DONT_MATCH_NAME` without using the results and then do another without and use the results

stacyharper:
> Maybe do a lookup with `DONT_MATCH_NAME` without using the results and then do another without and use the results

I would also recommend doing both, at least for some time until all config has been bumped.

GeopJr:
I think I found the best middle ground by checking the label and the `xdg:schema` if it's available.

Could both of you test #742 when possible?

stacyharper:
Nak!

stacyharper:
Oh wait. Actually it seems to works. Even after a fresh reboot. Maybe the update of the schema cause a failure on the first try?

GeopJr:
Looks like a different issue (at worst #742 would return 0 accounts). If you manage to reproduce it again, the actual error message should be a critical in logs ("Error while searching for items in the secret service: "). Usually it's portals related, not sure if the change back to DONT_MATCH_NAME caused it :/

PR #702, opened 2023-12-14
fix(ProfileCover): add separators

TheEvilSkeleton:
image


GeopJr:
LGTM, thanks!

Is there a particular reason for porting it to AdwPreferencesRow? Considering we don't really use its properties (title, markup), I'd say ListBoxRow is more appropriate

TheEvilSkeleton:
Uh yeah, good point

GeopJr:
Merging, thanks!

PR #712, opened 2023-12-19
feat(MediaViewer): add alternative shortcuts for zoom

oscfdezdz:


GeopJr:
LGTM Thanks!

Should it be both instead? Browsers seem to support both <kbd>Ctrl</kbd><kbd>_</kbd> <kbd>Ctrl</kbd><kbd>-</kbd> and <kbd>Ctrl</kbd><kbd>+</kbd> <kbd>Ctrl</kbd><kbd>=</kbd>

oscfdezdz:
<kbd>Ctrl</kbd><kbd>_</kbd> and <kbd>Ctrl</kbd><kbd>=</kbd> don't work for me, maybe it's a layout issue. There is `Gdk.Key.undebar`, I went with the `Gdk.Key.underscore` but as I can't test them, I'm not sure if there are differences.

GeopJr:
(the problem was that it only allowed the ctrl mask, I added a second statement to allow both ctrl and ctrl+shfit)

PR #718, opened 2023-12-24
fix(Base): narrow mode should always have a back arrow

GeopJr:
fix: https://github.com/GeopJr/Tuba/pull/648#issuecomment-1868146320
fix: #722


GeopJr:
rel: https://gitlab.gnome.org/GNOME/libadwaita/-/issues/775

PR #728, opened 2023-12-31
feat: Support zooming in Media Viewer via ctrl + mouse wheel (#727)

### LINKED ISSUE #727: [Request]: Zoom media via ctrl + mouse wheel
### Describe the request

Currently, zooming media seems to be only possible via pinch gestures and keys. However, it would be nice if it was also possible to zoom in/out via ctrl + mouse wheel, similarly to how it works in web browsers.

### Implementation Details

- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

mgorny:
Support zooming the Media Viewer via ctrl + mouse wheel scrolling. The controlled needed to be added straight into Gtk.ScrolledWindow, as otherwise it stops working when the media is zoomed in.

Closes #727

### DIFF
--- a/src/Views/MediaViewer.vala
+++ b/src/Views/MediaViewer.vala
@@ -238,6 +238,19 @@ public class Tuba.Views.MediaViewer : Gtk.Widget, Gtk.Buildable, Adw.Swipeable {
 			is_done = true;
 		}
 
+		private bool on_scroll (Gtk.EventControllerScroll scroll, double dx, double dy) {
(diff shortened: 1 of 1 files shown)


GeopJr:
(I'm not ignoring this but it has to wait for after the 0.6 patching period is done)

mgorny:
No problem. I've pushed the style fix.

GeopJr:
LGTM, thanks!

PR #730, opened 2023-12-31
fix: Make Tuba.Widgets.Avatar non-focusable (#729)

### LINKED ISSUE #729: [Bug]: The "avatar" button occurs twice in tab cycling, only one resulting in a functional menu
### Describe the bug

When tab cycling through window elements, the "avatar" button is cycled twice. When cycled the first time, enter opens the menu with account selection. When cycled the second time, enter doesn't do anything.

### Steps To Reproduce

1. Open Tuba.
2. Press Tab twice. "Avatar" button should become highlighted for the second time.
3. Press Enter. The menu shouldn't open.

### Logs and/or Screenshots

https://github.com/GeopJr/Tuba/assets/110765/bba9dac2-a7b1-4743-b89a-07988288791c

The sequence in this movie is:
1. Tab — selects the "avatar" button.
2. Enter — opens the menu.
3. Escape — closes the menu.
4. Tab — selects the "avatar" button the second time (you can notice the outline flashing).
5. Enter — only causes the button to flash without doing anything.
6. Tab — selects the hamburger button.

### Instance Backend

Mastodon

### Operating System

Gentoo Linux amd64

### Package

OS repositories

os: Gentoo (null)
prefix: /usr
flatpak: false
version: 0.6.1 (production)
gtk: 4.12.4 (4.12.4)
libadwaita: 1.4.2 (1.4.2)
libsoup: 3.4.4 (3.4.4)
libgtksourceview: 5.10.0 (5.10.0)

### Additional Context

_No response_

mgorny:
Make Tuba.Widgets.Avatar non-focusable to prevent the child button of accounts selector from being selected during tab-cycling.  It is duplicate to the accounts selector itself.

Closes #729

### DIFF
--- a/data/ui/views/sidebar/view.ui
+++ b/data/ui/views/sidebar/view.ui
@@ -39,6 +39,7 @@
 								<property name="popover">account_switcher_popover_menu</property>
 								<child>
 									<object class="TubaWidgetsAvatar" id="accounts_button_avi">
+										<property name="can_focus">false</property>
 										<property name="size">26</property>
 									</object>
 								</child>


GeopJr:
Nice catch!

I'd rather Widgets.Avatar stays focusable. Its proper use is to act as a button, apart from the sidebar headerbar where it's in a MenuButton (and causes the double-focus). Instead setting that specific instance's can-focus to false sounds more appropriate to me, what do you think?

https://github.com/GeopJr/Tuba/blob/8a4708d699e734d17305e9b08c08e5d7c32a1ad0/data/ui/views/sidebar/view.ui#L41-L43

mgorny:
Thanks for the suggestion. I've updated the PR and confirmed that this variant also works.

GeopJr:
LGTM, thanks!

Issue #734, opened 2024-01-02
[Bug]: Headerbar conditions regression


IBBoard:
### Describe the bug

#575 has come back - Tuba isn't showing the view switcher at narrow-but-not-narrow-enough-to-put-the-buttons-at-the-bottom sizes. It appears to be due to #651 fixing race conditions but breaking the initial value.

### Steps To Reproduce

1. Open Tuba with a default width of 555px
  * Expected: Title bar contains view switcher
  * Actual: Title bar contains title.
1. Resize Tuba below 550px
2. Resize Tuba above 550px
3. View switcher is shown

### Instance Backend

Mastodon

### Operating System

openSuse Tumbleweed

### Package

Compiled manually

### Troubleshooting information

```
os: openSUSE Tumbleweed (null)
prefix: /usr
flatpak: false
version: -6793a3c (development)
gtk: 4.12.4 (4.12.4)
libadwaita: 1.4.2 (1.4.2)
libsoup: 3.4.4 (3.4.4)
libgtksourceview: 5.10.0 (5.10.0)
```

### Additional Context

Built 8038e8f - broken.
Built 6793a3c (previous commit, before #651 merged) - correct behaviour.
Tried manually adding some of the old hooks back without any luck.
Added `info()` calls to `TabbedBaseView.title_stack_page_visible` but it doesn't log until the resize. So the breakpoint appears not to be handling the default size?

PR #735, opened 2024-01-02
fix(Main): switcher state and breakpoints between is_mobile

GeopJr:
fix: #734 

To avoid a breakpoint condition hell, let's just wrap it in another stack.

The problem:
We want the main window to show the main view switcher and switcher bar ONLY in narrow mode (`is-mobile`). To achieve the bottom bar one, is-mobile is bound to its visibility and the main view breakpoint is bound to its reveal property. That way, there are two visibility states. That is not the case for the switcher however, if the switcher is not visible, the headerbar stack cannot set it as the visible child, so both the breakpoint condition and the `is-mobile` one need to be in sync. That led to the previous *very* complicated solution.

This time let's just have 2 visibility states for it, like the switcher bar, using two stacks, one for `is-mobile` and one for the breakpoint.

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -30,6 +30,7 @@ namespace Tuba {
 
 		public Dialogs.MainWindow? main_window { get; set; }
 		public Dialogs.NewAccount? add_account_window { get; set; }
+		public bool is_mobile { get; set; default=false; }
 
 		public Locales app_locales { get; construct set; }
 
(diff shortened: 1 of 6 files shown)


GeopJr:
Merging and hoping for the best

IBBoard:
Works for me on an initial test. It started with the view buttons and they're appearing and disappearing as expected so far 🙂

PR #742, opened 2024-01-09
fix(SecretAccountStore): workaround DONT_MATCH_NAME

GeopJr:
fix: #701 #114 again 

# What's did #701 do?

When searching for secrets (saved accounts) in the keyring, we encountered issues with the keyring not unlocking in some setups. The solution was to retrieve ALL secrets that match the schema and not just Tuba's (#134, #114). However when working on another app that used the same schema as Tuba's I noticed that the two were reading each other's secrets and failing to load them (as expected) so I reverted the change on #701. Now there are two problems, both #114 and the one just described.

# What does this PR do?

It reverts #701 but to prevent the new problem, it checks if the label is "Mastodon Account" (not so reliable) and if `xdg:schema` matches the app id (`dev.geopjr.Tuba`) and skips the secret if it doesn't.

### DIFF
--- a/src/Services/Accounts/SecretAccountStore.vala
+++ b/src/Services/Accounts/SecretAccountStore.vala
@@ -13,7 +13,7 @@ public class Tuba.SecretAccountStore : AccountStore {
 		schema_attributes["version"] = Secret.SchemaAttributeType.STRING;
 		schema = new Secret.Schema.newv (
 			Build.DOMAIN,
-			Secret.SchemaFlags.NONE,
+			Secret.SchemaFlags.DONT_MATCH_NAME,
 			schema_attributes
 		);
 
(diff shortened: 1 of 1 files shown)


LukaszH77:
I tried `ea8933a` and it asks me to log in every time and had to go back to previous commit.

GeopJr:
Could you help me debug it by running #743?

It will log some critical messages when you start tuba for every saved account, it will look like this, at the top:

```
(dev.geopjr.Tuba:12004): Tuba-CRITICAL **: 18:09:13.208: SecretAccountStore.vala:78: SECRET-DEBUG:
label: Mastodon Account
has_schema: true
schema: dev.geopjr.Tuba
domain: dev.geopjr.Tuba
skip: false
```

LukaszH77:
```
$ flatpak run dev.geopjr.Tuba

(dev.geopjr.Tuba:2): Tuba-CRITICAL **: 17:22:43.676: SecretAccountStore.vala:78: SECRET-DEBUG:
label: Konto Mastodon
has_schema: false
skip: true

(dev.geopjr.Tuba:2): Tuba-CRITICAL **: 17:22:43.676: SecretAccountStore.vala:78: SECRET-DEBUG:
label: Konto Mastodon
has_schema: false
skip: true
```

PR #766, opened 2024-02-03
feat(announcements): banner and placement

GeopJr:
This PR does the following
- Move announcements from the sidebar to the menu
- Show a banner when new announcements are available

The questions are:
- Should it even move to the menu? The idea was that it takes space in the sidebar for almost no reason - unless there are announcements it's probably useless 99% of the time. However the workflow of showing it when there are new announcements (with a badge like notifications) and hiding it when there are not, is not ideal because it will either be visible until the request to announcements completes and then disappear if there are none OR be hidden and appear if there are any. Both of these have a UI shift
- Should by clicking "View" also mark all announcements as read? I think so, but you can mark as read individual announcements too so I'm not sure what's more desirable 

[Screencast from 2024-02-03 19-22-11.webm](https://github.com/GeopJr/Tuba/assets/18014039/bb32fbd5-78fa-471d-8b11-245b7697e0b0)


LukaszH77:
> The idea was that it takes space in the sidebar for almost no reason - unless there are announcements it's probably useless 99% of the time.

Yep, pretty much. Is there a way (and is it HIG-compliant if so) to have announcement in the sidebar only when there are unread announcements? But then again, there would be no way to see them after marking as viewed...

GeopJr:
Not visible in the recording above but it was moved to the main menu (the one next to "Tuba"), and will be there whether or not there are unread announcements

> Is there a way (and is it HIG-compliant if so) to have announcement in the sidebar only when there are unread announcements?

Tobias also suggested that in the circle review but my concerns with that is the UI shift. I tried to explain why on the first point in the issue but it's not that clear:

Basically, a request needs to be done to get the announcements to see if there are any unread ones. While that is being done, we need to decide whether "Announcements" should be in the sidebar until then and disappear after the request is done and there are no announcements OR, the opposite, have it not be there initially and only appear in the sidebar if there are any announcements

The problem with both is that the sidebar height will change from the initial state to the one after the request is done. If a user is trying to click "Follow Requests" right when the announcements request finishes, they'll click "Announcements" if it appears then


(On that note, "Follow Requests" needs to move away too, since it's only useful when there are any follow requests". Another banner?)

Issue #786, opened 2024-02-18
[Request]: Report Posts


GeopJr:
### Describe the request

I've been avoiding implementing some actions due to their importance but in light of the recent spam attacks, reporting posts should be implemented.

### Implementation Details

- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

andypiper:
I just dropped by to make this exact request! Thanks for looking into this - please include the option to forward the report to the originating instance.

GeopJr:
[Screencast from 2024-02-19 04-46-17.webm](https://github.com/GeopJr/Tuba/assets/18014039/b75735dc-a7fc-43e1-a4d8-cd6d7df698f4)

PR #787, opened 2024-02-19
feat(dialogs): report

GeopJr:
fix: #786 

TODO:

- ~~Test on every single backend~~
- ~~Add support for reporting profiles on their own~~

[Screencast from 2024-02-19 04-46-17.webm](https://github.com/GeopJr/Tuba/assets/18014039/bc53bf37-8f8f-4b6a-a349-60f01a08f851)

### DIFF
--- a/data/ui/menus.ui
+++ b/data/ui/menus.ui
@@ -48,12 +48,6 @@
         </item>
         <section>
           <item>
-            <attribute name="label" translatable="yes">Report</attribute>
-            <attribute name="action">view.report</attribute>
-            <!-- TODO: Reporting users -->
-            <attribute name="hidden-when">action-missing</attribute>
-          </item>
-          <item>
             <attribute name="label" translatable="yes">Block</attribute>
             <attribute name="action">view.blocking</attribute>
           </item>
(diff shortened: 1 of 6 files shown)


GeopJr:
I did some testing on gotosocial and akkoma but wasn't too extensive as setting everything up is a bit too complex on my current setup. I'm a bit hesitant to merge it due to its criticality but everything should work as expected. In case any passerby noticed anything wrong with the workflow, let me know:

- form data
- account_id and forward is always provided
- if the content is not empty then it's provided
- default category is other
- if the category is violation, then rules are also provided
- status_ids are always provided if any

GeopJr:
If you want to test it, change this line https://github.com/GeopJr/Tuba/pull/787/files#diff-4882facb54040a914f20185443b33577aabeef26190156d8976ef2dd1a8c36fdR346 to `https://faraway.town/api/v1/reports` (it's my server, reports will be ignored as it's abandoned)

[Don't be like me](https://tech.lgbt/@GeopJr/111964023498426371), don't annoy your instance admins with bogus reports!!!

PR #800, opened 2024-02-24
feat: filters

GeopJr:
fix: #200 

~~Filter settings should be split into a separate FR~~ I'll do it here ig

There's stuff I'm unsure about and stuff I'd rather hadn't added:
- When a filter is set to hide completely, do we have to handle it at all? Shouldn't the server already handle that???
- When multiple filters apply to a single post, do we show all filter titles or only the first one?
- I'd rather we didn't have to use yet another stack. The status widgets are getting more and more bloated. There's now boxes in boxes in boxes in stacks in boxes and buttons and labels and flowboxes... it's getting ridiculous. Oh well :shrug:  

[Screencast from 2024-02-24 16-59-18.webm](https://github.com/GeopJr/Tuba/assets/18014039/95d7323b-a2e7-4d05-b047-99026401891c)


GeopJr:
I'll do the settings on this PR too I guess

:warning: Tuba will only support v2 filters. Don't worry, if your backend errors during the initial call, the page won't be visible. :warning:

TODO:
- Language filters
- Edit page
- Deletions
- Saving


I'm not sure about the layout :/ maybe the expander rows are a bit too much? (they still need padding ofc). Also I doubt there are keyword limits on all backends so I set it to only expand if there are at most 10 keywords. The edit dialog would still need to display all of them.

[Screencast from 2024-02-24 22-45-53.webm](https://github.com/GeopJr/Tuba/assets/18014039/ffe0344c-3ebc-4d00-b069-968e6a216976)

GeopJr:
Still so much to do but somewhat getting there. Not so fun fact, there are 0 limits on filters as far as I can tell. No keyword amount limit, not keyword length limit, no filter limit, no title length limit... :shrug: 

[Screencast from 2024-02-25 23-33-26.webm](https://github.com/GeopJr/Tuba/assets/18014039/116e4771-def1-4ad5-8428-1a131512de37)

PR #804, opened 2024-02-26
feat: port to adaptive dialogs

### LINKED ISSUE #777: [Request]: Move most of the dialogs to Adw.Dialog
### Describe the request

A new dialog concept, Bottom Sheets, will be added to the libadwaita library in version 1.5. This will require the use of a new widget, namely Adw.Dialog. On the libadwaita documentation page, you can find examples of how to use this feature and an example of how this widget looks on the screen. IMO, «Compose» should be migrated to this first ASAP.

Docs: https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.Dialog.html

### Implementation Details

- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

oscfdezdz:
All remaining dialogs, except for the composer (#623), should be covered by this PR.

Fixes #777


oscfdezdz:
Something is wrong, <kbd>Esc</kbd> key to close dialogs doesn't work and it should by default with AdwDialog.

GeopJr:
(I haven't looked into this PR yet)

Does it happen on all dialogs or some? In some cases we handle it https://github.com/search?q=repo%3AGeopJr%2FTuba+add_binding_action+%28Gdk.Key.Escape&type=code. Anything that is not set to window.close should probably have its AdwDialog's can-close set to false

oscfdezdz:
All, even with Report dialog that has `report.back` binding and not `window.close`. It seems like the main window is capturing the shortcut instead, if the active AdwNavigationPage has a back button it would go back when the dialog is open

From what I understand from the migration guide and code, we should not have to explicitly specify it https://gitlab.gnome.org/GNOME/libadwaita/-/blob/main/src/adw-dialog.c?ref_type=heads#L1188-L1189

GeopJr:
Yeah looks like the global esc accel is causing it https://github.com/GeopJr/Tuba/blob/354cde1016d9b1f200e2fb4667ec1e67052bf8d8/src/Application.vala#L251

Removing it (`"Escape"`) seems to fix it, though it's still needed :/ Maybe we can check if any dialogs are visible and handle it?

cc: @alice-mkh

alice-mkh:
Why do you even need that action? Navigation view already has an action for going back AND handles shortcuts

GeopJr:
The mediaviewer is an overlay and app.back dismisses it if it's open. I removed the ones navigationview handles and made the mediaviewer dismiss itself on its keypress event instead

alice-mkh:
Can't you use navigation view for media viewer?

src/Views/MediaViewer.vala
```diff
@@ -540,21 +540,37 @@ public class Tuba.Views.MediaViewer : Gtk.Widget, Gtk.Buildable, Adw.Swipeable {
 	protected bool on_keypress (uint keyval, uint keycode, Gdk.ModifierType state) {
```
alice-mkh, line 540: Wait. Why are you doing _any_ of this instead of using shortcuts? That's why it breaks.

alice-mkh:
I see. Well either way, the way it handles shortcuts is _very_ broken and things would work if you redo it the correct way, using `add_binding()` etc.

alice-mkh:
I do want to support transitions like in nav view in future, but yeah, right now it's not possible.

src/Views/MediaViewer.vala
GeopJr: They wouldn't get triggered IIRC (Escape didn't either when I tried a few hours ago) I'll take a closer look into it at some point :/

alice-mkh: Well, if things don't work you investigate why they don't work instead of replacing it with a hack that will break other things later...

It probably didn't work because the widget wasn't focused, or something like that

GeopJr: I didn't even consider it a hack to be honest 😭 I found out what I was doing wrong, setting up a ShortcutController fixed my issues. I'll open a separate PR, thanks for point it out!

alice-mkh: I mean this _was_ the way things worked in GTK3. Except even then you needed extra code to e.g. make sure non-en_US keyboard layouts didn't break shortcuts. (which your func doesn't have either)

Issue #153, opened 2023-04-04
[Request]: composer syntax highlighting 


GeopJr:
### Describe the request

- [X] markdown (only when supported)
- [X] html (only when supported)
- [X] highlight @<handle> #<hashtag> :emoji:

see: #137 

### Implementation Details

- [ ] This should be an option in settings.
- [X] This should be only available to some fediverse backends. (Include which ones on the above field).
- [X] This is client-only (and shouldn't sync with the instance).
- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

nekohayo:
In case this is helpful, you might want to take a look at GTG's code for handling highlighting of tags in semi-realtime (it waits a fraction of a second to parse stuff for [performance reasons](https://github.com/getting-things-gnome/gtg/issues/270) because we have a lot of costly operations happening in the backend and data format otherwise)

* [GTG/gtk/editor/taskview.py around line 45](https://github.com/getting-things-gnome/gtg/blob/HEAD/GTG/gtk/editor/taskview.py#L45) has a regex for finding tags etc.
* [GTG/gtk/editor/text_tags.py](https://github.com/getting-things-gnome/gtg/blob/HEAD/GTG/gtk/editor/text_tags.py) is probably / maybe where the main stuff happens for highlighting? I'm not sure
* [GTG/gtk/colors.py](https://github.com/getting-things-gnome/gtg/blob/HEAD/GTG/gtk/colors.py)

GeopJr:
Thanks for the info!

I'll take a closer look but I had mostly (ab)using gtksourceview's syntax highlighting in mind. Instead of parsing the input and highlighting manually, we could write a custom [lang file](https://gitlab.gnome.org/GNOME/gtksourceview/tree/master/data/language-specs) that handles @ # : and perhaps some markdown with asterisks, blockquotes, backticks, lists. Unsure if it's possible as I haven't looked further into it but I assume it will be more perfomant and safe than regex (considering the amount of rules)

PR #823, opened 2024-03-07
feat(Composer): syntax highlighting

GeopJr:
fix: #153 

Last piece of the composer puzzle, syntax highlighting

Took longer than expected, but it's here:

| Plain | Markdown | HTML |
| :---: | :---: | :---: |
| Screenshot from 2024-03-07 04-12-41 | Screenshot from 2024-03-07 04-12-53 | Screenshot from 2024-03-07 04-13-04 |


GeopJr:
Windows is mostly guesswork based on the zip :shrug:

Issue #865, opened 2024-03-20
[Bug]: focused item is lost when opening a post


WhyNotHugo:
### Describe the bug

When I open a post and return to the main screen, the previously focused post is lost; I have to start browsing from the top again.

### Steps To Reproduce

1. Press tab a few times until the first post in the timeline is focused.
2. Press the down arrow key to scroll through posts.
3. Press the space bar to open a focused post and see its replies.
4. Press Escape to return to Home.
5. The previously focused post is no longer focused. It seems that "nothing" has focus, and pressing tab once focuses the left-side menu. Pressing tab a few more times moves focus back to the Home area, but the very topmost item is focused (rather than the previously focused one). I now have to scroll through my Home from the topmost post again, losing my previous "position".

### Logs and/or Screenshots

n/a

### Instance Backend

Mastodon

### Operating System

Alpine Linux edge

### Package

OS repositories

### Troubleshooting information

os: Alpine Linux (null)
prefix: /usr
flatpak: false
version: 0.6.3 (production)
gtk: 4.14.0 (4.12.5)
libadwaita: 1.4.4 (1.4.2)
libsoup: 3.4.4 (3.4.4)
libgtksourceview: 5.10.0 (5.10.0)

PR #866, opened 2024-03-21
feat(main_window): keep track of last focused item between navigation changes

GeopJr:
fix: #865 

needs memory leak check

a bit weird way to do this but it's the best non-listbox-specific way I could think of

Fixed the focus on media viewer toggling too since it didn't work as expected


GeopJr:
This is technically a feature and shouldn't be included in the patch, however, not keeping track of focus between navigation changes should be considered a bug so I'll include it.

PR #876, opened 2024-03-28
feat: v0.7.1

GeopJr:
```
# Fixed

- Syntax highlighting not working if there are no available content types (#864)
- Keyboard navigation focus not being remembered between NavigationView page changes and MediaViewer reveal toggling (#866)
- Changed brand colors to ones with more contrast (#867, thanks [at]bertob)
- MediaViewer videos playing before being initialized and MediaViewer buttons not being updated to support the video layout due to a race condition (#869)
- MediaViewer scale revealer dismiss widget updating only when gestures are used instead of page changes, causing it to show a different widget when dismissing without using gestures (#872)
- Post audio attachments not having an overlay icon background when there's a thumbnail available (#873)

# Package maintainers

- It was omitted from the last changelog, but since v0.7.0, Tuba depends on `icu-uc` for extended character counting

**Full Changelog**: https://github.com/GeopJr/Tuba/compare/v0.7.0...v0.7.1
```

### DIFF
--- a/build-aux/dev.geopjr.Tuba.Devel.json
+++ b/build-aux/dev.geopjr.Tuba.Devel.json
@@ -34,7 +34,7 @@
             "sources" : [
                 {
                     "type" : "git",
-                    "url" : "https://gitlab.gnome.org/chergert/libspelling.git",
+                    "url" : "https://gitlab.gnome.org/GNOME/libspelling.git",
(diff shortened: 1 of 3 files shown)


GeopJr:
self reminder: change owner name of libspelling on flathub and bump to latest tag

PR #884, opened 2024-04-02
ui: Fix translatable usage

yakushabb:
Marking translatable="yes" or translatable="1" actually achieves the same result, but almost all projects use the first one.

It's good to follow common practices.


GeopJr:
Sounds good to me, thanks!

Issue #885, opened 2024-04-02
[Bug]: fractional scaling font blurriness on 0.7


alexreid:
### Describe the bug

When using a 125% fractional scaling factor on Gnome, Tuba 0.7.1 has rendering issues and blurry text compared to Tuba 0.6.3.

### Steps To Reproduce

1. Install Tuba 0.7.1
2. Set display scale to 125% in Gnome Settings (you may need to enable this via `gsettings set org.gnome.mutter experimental-features "['scale-monitor-framebuffer']"` depending on distro)
2. Open Tuba
3. Note that text looks blurrier, and scrolling can produce strange graphical artefacts

### Logs and/or Screenshots

0.6.3:

![Screenshot from 2024-04-02 22-58-05](https://github.com/GeopJr/Tuba/assets/2099579/5e234246-ba53-42e5-a561-984ef4a575da)

0.7.1:

![Screenshot from 2024-04-02 22-58-55](https://github.com/GeopJr/Tuba/assets/2099579/59a2f96b-b9c7-40fe-8805-5843a0677646)

### Instance Backend

Mastodon

### Operating System

Fedora Workstation 39, Gnome Shell 45.5, Wayland session

### Package

Flatpak

### Troubleshooting information

```
os: GNOME 46 (Flatpak runtime)
prefix: /app
flatpak: true
version: 0.7.1 (production)
gtk: 4.14.1 (4.14.1)
libadwaita: 1.5.0 (1.5.0)
libsoup: 3.4.4 (3.4.4)
libgtksourceview: 5.12.0 (5.12.0)
```

<details>

<summary>Troubleshooting info from 0.6.3, installed from Fedora Flatpak repo</summary>

```
os: Fedora Linux 39 (Thirty Nine)
prefix: /app
flatpak: true
version: 0.6.3 (production)
gtk: 4.12.5 (4.12.5)
libadwaita: 1.4.3 (1.4.3)
libsoup: 3.4.4 (3.4.4)
libgtksourceview: 5.10.0 (5.10.0)
```

</details>

GeopJr:
Thanks for the report!

Could you try `GSK_RENDERER=ngl flatpak run dev.geopjr.Tuba` and tell me? I see some blurriness here but it's far less noticeable than on your screenshot

[...]

PR #886, opened 2024-04-04
fix: fractional scaling blurriness with gl

GeopJr:
fix: #885 

On 46, GTK defaults to ngl as its renderer. While it's a nice step forward, performance is not on par with gl yet and is very noticeable on Tuba, so we forced `gl` as the default on #847 until everything gets ironed out (gl is being used only if nothing else is set, so you can still override it).

On #885 an issue with fractional scaling and blurriness was raised and after researching on GTK's issue tracker, I discovered that setting `GDK_DEBUG=gl-no-fractional` reproduced the old behavior. This PR sets that if the renderer is gl.

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -201,7 +201,11 @@ namespace Tuba {
 				if (GLib.Environment.get_variable ("SECRET_BACKEND") == "file")
 					GLib.Environment.set_variable ("SECRET_FILE_TEST_PASSWORD", @"$(GLib.Environment.get_user_name ())$(Build.DOMAIN)", false);
 			#endif
+
 			GLib.Environment.set_variable ("GSK_RENDERER", "gl", false);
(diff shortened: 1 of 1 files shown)


thibaultamartin:
Would you consider doing a minor release solely for this fix? It might not be much in terms of code, but the impact is pretty significant for end users :)

GeopJr:
It will be a patch! But I'm holding off until the end of the week in case anything else critical or a translation fix comes up

In the meantime, if you are using the flatpak, you can run `flatpak override --env=GDK_DEBUG=gl-no-fractional dev.geopjr.Tuba` to override it

(though it might cause issues when we move to ngl eventually)

GeopJr:
(Btw, Image Viewer (Loupe), also uses `gl` instead of `ngl` and suffered from the same thing. I [echoed this issue there](https://gitlab.gnome.org/GNOME/loupe/-/issues/346) and the fix will be available on the next release)

Issue #895, opened 2024-04-10
[Bug]: Tuba crashes when trying to create new post or reply on Pixelfed


LukaszH77:
### Describe the bug

It used to work, but now when I'm trying to create a new post on pixelfed.social or to reply to someone Tuba crashes. I think it started when they introduced this change: https://mastodon.social/@pixelfed/112222959196646449

On another Pixelfed server, one with previous version, posting from Tuba still works.

### Steps To Reproduce

1. Login to pixelfed.social
2. Try to create a new post or to reply
3. Tuba crashes

### Logs and/or Screenshots

```
<paste your logs here>
```

### Instance Backend

I'm not sure / Other

### Operating System

openSUSE Tumbleweed

### Package

Flatpak

### Troubleshooting information

os: GNOME 46 (Flatpak runtime)
prefix: /app
flatpak: true
version: main-ce9a6db (development)
gtk: 4.13.9 (4.13.9)
libadwaita: 1.5.0 (1.5.0)
libsoup: 3.4.4 (3.4.4)
libgtksourceview: 5.11.2 (5.11.2)

GeopJr:
Do you have any logs of the crash? For me, on pixelfed.social, it shows an error instead (which I'm debugging rn)

GeopJr:
I just crashed due to `https://pixelfed.social/api/v1/custom_emojis`

GeopJr:
Attempted a fix at #896, let me know if it works for you too!

LukaszH77:
It works! Thank you!

PR #896, opened 2024-04-10
fix(composer): pixelfed posting

GeopJr:
fix: #895 

The first commit will fix it but I'm not sure if the second should be merged.

The first issue is that Pixelfed for some reason broke their custom emojis api but also Vala's optional chaining not working as always. So it could be null and still return as truthy.

The second issue, which was just critical logs from json glib but should be taken into account, was that when attempting to get the json root, it could *not* be an object, as many edge cases are valid json but useless to us.

It makes sense to check but is also useless to assume that a server could *not* return an object. Well, if they implement mastoapi that is.


GeopJr:
Reverted the second part for now.

Issue #905, opened 2024-04-13
[Bug]: Custom emojis do not work in profile fields


david-swift:
### Describe the bug

Custom emojis are not displayed as emojis in the labels of fields of a profile.

### Steps To Reproduce

View a Mastodon profile with custom emojis in the fields (e.g. [@ErikUden@mastodon.de](https://mastodon.de/@ErikUden)). 

### Logs and/or Screenshots

Tuba:
![Screenshot from 2024-04-13 14-36-08](https://github.com/GeopJr/Tuba/assets/106754840/d6242bc8-f3a7-4d52-aafa-f4e2508bf5a5)

Web view:
![Screenshot from 2024-04-13 14-38-11](https://github.com/GeopJr/Tuba/assets/106754840/23368645-fa94-4c5c-a7f1-63b598b09ce5)

### Instance Backend

Mastodon

### Operating System

Fedora 40

### Package

Flatpak

GeopJr:
Nice catch! Took the opportunity to redesign it slightly so they are aligned vertically as some values can be way too long

![Screenshot from 2024-04-13 17-41-41](https://github.com/GeopJr/Tuba/assets/18014039/abac82b9-5ff4-4f13-b16d-c82a3dc1dfd8)

david-swift:
Nice! I really like that they are selectable now! The "Scale Custom Emojis on Hover" option doesn't seem to work with the fields currently though.

I played around with the design a bit and IMO, the fields would look better and more GNOMEy with a property row design instead of making both the label and value having the same weight.

![Screenshot from 2024-04-13 18-22-13](https://github.com/GeopJr/Tuba/assets/106754840/25ae7629-e1fc-44ee-9bce-925ac36c4e7e)

david-swift:
![Screenshot from 2024-04-13 18-26-58](https://github.com/GeopJr/Tuba/assets/106754840/3d442519-8cbd-4118-b542-71387f046dd0)

david-swift:
@GeopJr to make sure you receive a notification :)

GeopJr:
Sounds good to me!

Some follow ups:

- Do we make it dim label? Both unicode and custom emojis will be greyed out. The alternative is setting text color to grey-ish but that might have other a11y issues with high contrast and whatnot
- Do we place the checkmark in the start or the end?

![image](https://github.com/GeopJr/Tuba/assets/18014039/3693ca4b-61e8-4ac3-b777-1d9c9fd61680)

[...]

PR #907, opened 2024-04-13
feat(profile-cover): fields design

GeopJr:
fix: #905

TODO:
- edge cases
- leak test


GeopJr:
fixed edge cases using a size group
leak test passed

Issue #397, opened 2023-07-19
[Bug]: Light image viewer


bertob:
### Describe the bug

When opening an image the image viewer is no longer dark, which is not in line with the HIG.

It seems that this was motivated because the floating buttons were not visible enough (?), but it's definitely not a good solution. For an app like this keyboard shortcuts / Ctrl + scroll / pinch to zoom would probably be enough to be able to just drop the explicit zoom buttons.

### Steps To Reproduce

1. Open an image in the timeline

### Logs and/or Screenshots

![image](https://github.com/GeopJr/Tuba/assets/1908896/791dee7e-79fd-4a43-a95b-0797cbb7a501)

### Instance Backend

Mastodon

### Operating System

Fedora 38

### Package

Flatpak

### Troubleshooting information

os: GNOME 44 (Flatpak runtime)
prefix: /app
flatpak: true
version: 0.4.0 (production)
gtk: 4.10.3 (4.10.3)
libadwaita: 1.3.2 (1.3.2)
libsoup: 3.4.2 (3.4.2)
libgtksourceview: 5.8.0 (5.8.0)

GeopJr:
Sounds good to me!

However, that issue will probably come up again for the pagination buttons, should they also be removed?

note to self: add <kbd>Ctrl</kbd><kbd>+</kbd>, <kbd>Ctrl</kbd><kbd>-</kbd> & <kbd>Ctrl</kbd> + mouse wheel shortcuts

Issue #299, opened 2023-05-30
[Request]: Better video controls


bertob:
### Describe the request

![image](https://github.com/GeopJr/Tuba/assets/1908896/834cd95c-f22c-4cac-8181-4e9e7e8abdee)

The video controls could be better, in particular:
- The huge play icon in the center looks clickable, but it isn't. I'd either make it a button or drop it
- The video should probably auto-play when you open the media viewer
- The overlay controls at the bottom are tiny and feel cramped. I'd give them more padding (e.g. 12px all around). Inspector hack:

![image](https://github.com/GeopJr/Tuba/assets/1908896/94572a97-808b-4231-9f99-0f7aff0b2203)

### Implementation Details

- [ ] This should be an option in settings.
- [ ] This should be only available to some fediverse backends. (Include which ones on the above field).
- [ ] This is client-only (and shouldn't sync with the instance).
- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

GeopJr:
> The huge play icon in the center looks clickable, but it isn't. I'd either make it a button or drop it

Part of GtkVideo (#205, https://gitlab.gnome.org/GNOME/gtk/-/issues/4522 (seems to be waiting designer input))

> The video should probably auto-play when you open the media viewer

✅

> The overlay controls at the bottom are tiny and feel cramped. I'd give them more padding (e.g. 12px all around)

Also part of GtkVideo, GtkMediaControls, but can be disabled and replaced.

Since Fractal is also using GtkVideo with GtkMediaControls I feel like it might be better to look for a common solution either for libadwaita or gtk directly

GeopJr:
The first issue has been fixed on GTK and it will be available in 4.11.4!

For the media controls I'll override its style

Thanks!

PR #909, opened 2024-04-14
feat(MediaViewer): enable GraphicsOffload for videos

rmader:
In order to allow pass-through of video content to Wayland compositors, reducing copies on the GPU and thus improving battery life and performance.

This change should generally be safe to do, however given the higher chances of triggering driver bugs good testing is due.

---

An easy way to test this is to look at `intel_gpu_top`/`radeontop`/ similar tools - in many cases this should be able to reduce GPU work considerably, especially for high-res content.

Note that for now it only affect hardware-decoded content - this will hopefully change in the future.

Draft because:
 - [ ] More testing is needed
 - [ ] Some possible design changes to e.g. hide the top bar in fullscreen mode after a timeout would increase the effect even more. This might be worth to investigate before merging. Doing this with the GTK inspector allows me to get zero-copy playback, which especially on mobile and low-end devices is really good.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,8 +46,8 @@ jobs:
         manifest-path: "build-aux/dev.geopjr.Tuba.Devel.json"
         cache-key: flatpak-builder-${{ github.sha }}
         arch: ${{ matrix.arch }}
-        repository-name: flathub-beta
-        repository-url: https://flathub.org/beta-repo/flathub-beta.flatpakrepo # until GNOME 45
+        repository-name: flathub
+        repository-url: https://dl.flathub.org/repo/flathub.flatpakrepo
   snap-builder:
     name: "Snap Builder"
     runs-on: ubuntu-latest
--- a/build-aux/dev.geopjr.Tuba.Devel.json
+++ b/build-aux/dev.geopjr.Tuba.Devel.json
@@ -1,7 +1,7 @@
 {
     "app-id": "dev.geopjr.Tuba",
     "runtime": "org.gnome.Platform",
-    "runtime-version": "46beta",
+    "runtime-version": "46",
     "sdk": "org.gnome.Sdk",
     "command": "dev.geopjr.Tuba",
     "finish-args": [
--- a/src/Views/MediaViewer.vala
+++ b/src/Views/MediaViewer.vala
@@ -826,6 +826,9 @@ public class Tuba.Views.MediaViewer : Gtk.Widget, Gtk.Buildable, Adw.Swipeable {
 
 		if (media_type.is_video ()) {
 			var video = new Gtk.Video ();
+
+			video.graphics_offload = Gtk.GraphicsOffloadEnabled.ENABLED;
+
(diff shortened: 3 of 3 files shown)


rmader:
@GeopJr: just to ask in advance: would you be open to make the video top-bar auto-hide together with other video controls in fullscreen mode? I think that would be good both from a design and a performance perspective :)

On touch devices all controls would probably need to toggle on/off on tap.

GeopJr:
Yeah, I wouldn't mind removing the headerbar on fullscreen altogether and instead have a floating 'toggle fullscreen' button (visible on mouse movement/click and such)!

In general, I'll follow anything the design team suggests. In the past, Tobias on #299 suggested making the state overlay of GtkVideo clickable (done upstream) and adding padding around the media controls (ignored upstream, done here). On #397, dropping the zoom buttons was mentioned but no further discussion took place

As I mentioned briefly on #299, it might be time for a libadwaita video player or controls so other apps like Fractal can also benefit from it - rather than designing and implementing one just for Tuba

rmader:
With further changes in the pipe - like https://github.com/GeopJr/Tuba/pull/913 - I think this is good to go. Tested on a few devices and haven't seen any regression - while there are some pretty nice performance improvements in various cases.

FTR: I think this is also good for a point release. In case some people see regressions, we'd most likely have to fix them on the driver side. Tuba is one of the first apps to (hopefully :P) adopt this new feature, but fortunately would be much less badly affected than e.g. a video player. I.e. if people using specific hardware would see buggy output when watching videos, that wouldn't break the main features of the app.

GeopJr:
Thanks for leading this!

Btw, I don't know if it makes a difference, but FWIW, Tuba does not use `ngl`, it uses `gl` (due to some performance regressions). It won't override `GSK_RENDERER` if manually set however

Feel free to experiment on Tuba if needed in the future again! It's still 'alpha' software so I don't mind if it's used as a testing ground

rmader:
Oh, the Flatpak build seems to fail because it's still using GTK `4.13.9`, but this needs `4.14` - do you happen to know where to fix that?

Regarding `ngl` and `gl` - fortunately the offloading is still supported with the old renderer, allowing it to work on devices without GLES 3.0 support. There are not many of them any more, but some mobile devices like the Librem5 are among them - and that benefits a lot from the offloading, more than most more powerful devices.

> Feel free to experiment on Tuba if needed in the future again! It's still 'alpha' software so I don't mind if it's used as a testing ground

Thanks a lot for the app, the help and the quick responses! I use Tuba as my main Mastodon client now and am really happy with it already - and will surely come back from time to time with issues or ideas.

GeopJr:
> do you happen to know where to fix that?

I think `46` does use 4.14 but `46beta` doesn't (and Tuba should switch to `46` now that it's released I guess). I'll do it don't worry about it. (The runtimes are maintained on https://gitlab.gnome.org/GNOME/gnome-build-meta)

rmader:
Just wanted to play with the flatpak and stumbled over the 46beta entry there - with that, I suppose the CI should succeed now.

rmader:
Yay, it worked.

PR #913, opened 2024-04-16
feat(MV): remove headerbar in fullscreen and reveal fabs on single click

GeopJr:
~~TODO: figure out why the motion event doesnt get triggered on fullscreen when the item at position is a video~~

Fixed, turns out when a video is playing the even controller motion's motion signal gets triggered constantly even though the cursor did not actually move.


rmader:
Thanks, this looks great! Gave it a try and see one more issue: the fullscreen toggle button doesn't reliably automatically hide with the other window controls yet - and when it does, it seems to do it with a delay.

rmader:
@GeopJr I'm thinking about a follow-up change to toggle fullscreen mode automatically - e.g. when the window is already maximized. Would need feedback from design, but what do you think about it?

For my usage patterns that would work quite well, especially as hitting the "back" button brings one back. AFAIK this doesn't quite work for swipe back yet, but that is probably an easy fix as well.

GeopJr:
> the fullscreen toggle button doesn't reliably automatically hide with the other window controls yet - and when it does, it seems to do it with a delay.

Noticed that too but can't figure out where exactly the issue is without digging deeper. It works exactly like GtkVideo's for the media controls (`EventControllerMotion` & a 5s timeout) but for some reason it doesn't get triggered when the current item is a video, only on full screen :shrug: needs further debugging, though I do have some suspicions

> I'm thinking about a follow-up change to toggle fullscreen mode automatically - e.g. when the window is already maximized. Would need feedback from design, but what do you think about it?

Main objection from my side is that fullscreen kind-of 'locks you'. (AFAIK) notifications wont come through and the hot corner is not active. It's up to the design team, but I'm okay with adding a gsettings option if needed

However, if this becomes default, then I don't think this PR is the best option as the menu is not visible. I'd either opt for showing the headerbar on motion/touch or adding another floating button for the menu

> AFAIK this doesn't quite work for swipe back yet, but that is probably an easy fix as well.

I didn't notice anything wrong with the dismiss gesture, is there anything I missed? (FWIW, the dismiss gesture is two finger vertical swipe)

rmader:
> Main objection from my side is that fullscreen kind-of 'locks you'. (AFAIK) notifications wont come through and the hot corner is not active. It's up to the design team, but I'm okay with adding a gsettings option if needed
> However, if this becomes default, then I don't think this PR is the best option as the menu is not visible. I'd either opt for showing the headerbar on motion/touch or adding another floating button for the menu

Yeah, it would be a quite unusual change - definitely not for this MR, rather something I'd like to play with. And it would definitely need design team review.

> I didn't notice anything wrong with the dismiss gesture, is there anything I missed? (FWIW, the dismiss gesture is two finger vertical swipe)

The swipe usually does not seem to get me back to the overview(?), unlike the mouse back button. But that's a minor issue not really related to this MR and I should report it independently :)

Issue #916, opened 2024-04-20
[Bug]: "Follow" button sensitivity does not match Mastodon's web UI / does not reflect blocked status?


nekohayo:
### Describe the bug

If I am not mistaken, the profile viewer in Tuba does not reflect blocked status.

Mastodon reflects that with the "Follow" button being disabled (and being unable to see someone's toots while their profile is generally public), I believe. Maybe Tuba should do the same, or, if you want the UI to be extra clear, show some sort of infobar or label that says that the user has probably blocked you from following or viewing their posts with your account.

### Steps To Reproduce

1. View the profile of a user who has (presumably) blocked you
2. Compare the "Follow" button in Tuba vs the Mastodon web interface

### Logs and/or Screenshots

As you can see in the example below, the web interface (shown on the left) seems to hint that I have been blocked by this person (I have no other explanation, as we were mutual followers and I have never muted nor blocked them, so it can't be something on my end), but Tuba's "Follow" button (as visible on the right in the video) seems to be shown as an normally interactive/sensitive widget:

[follow button sensitivity out of sync.webm](https://github.com/GeopJr/Tuba/assets/479401/e84a006d-d866-48d9-8b98-b2216bdf7474)

I have not tried clicking it to see what would happen.

### Instance Backend

Mastodon

### Operating System

Fedora 39

### Package

Flatpak

### Troubleshooting information

os: GNOME 46 (Flatpak runtime)
prefix: /app
flatpak: true
version: 0.7.2 (production)
gtk: 4.14.1 (4.14.1)
libadwaita: 1.5.0 (1.5.0)
libsoup: 3.4.4 (3.4.4)
libgtksourceview: 5.12.0 (5.12.0)

GeopJr:
Nice catch, dont know how we missed this lol

![Screenshot from 2024-04-21 18-43-31](https://github.com/GeopJr/Tuba/assets/18014039/ee9c299e-4f1b-4f7c-a240-633dfdb5109c)

PR #919, opened 2024-04-21
fix(relationship): blocked_by

GeopJr:
Don't know how we missed this lol

fix: #916

### DIFF
--- a/src/API/Relationship.vala
+++ b/src/API/Relationship.vala
@@ -25,7 +25,8 @@ public class Tuba.API.Relationship : Entity {
 		else if (followed_by)
 			label = _("Follows you");
 		else if (blocked_by)
-			label = _("Blocked");
+			// translators: as in, you've been blocked by them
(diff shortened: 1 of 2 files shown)


nekohayo:
It might be better to be more specific with _"Blocked you"_ instead of just "Blocked", because it could be mistaken and lead users to think, "Wait, I blocked them? I can't unblock them, what is this I don't even—".

Then it could have a GtkTooltip on hovering that label (and the follow button), to explain the consequences (i.e. you can't see their posts with your current account, and you can't follow them).

GeopJr:
+1 for "Blocked you" or "Blocks you"

unsure about the tooltip, it'd be too long. IMO being blocked implies that you cant interact with them, just like mutuals implies that they also follow you. Not opposed to giving more info but I don't think tooltip is the best way to do it(?). Maybe if it was something shorter "You cannot interact with them anymore"

Issue #924, opened 2024-04-23
[Bug]: Frequent crashes when playing high-res videos


rmader:
### Describe the bug

I made a post with 4k@60fps test videos here: https://floss.social/@rmader/112320017496066938

Right now playing these does not work well. If it works it takes long to start - and in many cases it doesn't start at all or crashes for me. This seems to mainly have to do with the download process: once the videos where downloaded successfully, things seem to work reliably. 

These are likely Gstreamer etc. issues, but I wanted to have a tracker bug here as well to refer to.

### Steps To Reproduce

1. find and open the post at https://floss.social/@rmader/112320017496066938
2. try to play the videos

### Instance Backend

I'm not sure / Other

### Operating System

Linux

### Package

Flatpak

### Troubleshooting information

Tested with Flatpak and manually compiled from `main`.

GeopJr:
I haven't looked into this yet but for completion:
- The download process is on https://github.com/GeopJr/Tuba/blob/main/src/Utils/Host.vala#L37-L71
- And the path gets set as the GtkVideo's filename

Ideally, I'd rather we *streamed* these rather than downloading them and then playing them but it's blocked by https://gitlab.gnome.org/GNOME/gtk/-/issues/4062 which is blocked by https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/724

rmader:
> Ideally, I'd rather we streamed these rather than downloading them and then playing them but it's blocked by https://gitlab.gnome.org/GNOME/gtk/-/issues/4062 which is blocked by https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/724

Ah, that's great to know, will put that on my list to look into.

Rafostar:
Hi, I dunno how Tuba would feel about using 3-rd party stuff, but it could technically use/integrate [Clapper](https://github.com/Rafostar/clapper) in its UI for playback, which:

a) Can actually stream video like this without the need to download (which seems inefficient for social media client)
b) Can stream with working seeking `Peertube` videos (both HLS and MP4 - your issue #662) - you just have to replace `https://` URI scheme with `peertube://` to let it understand that this website hosts peertube
c) Has `VALA` bindings support (working example/test app [here](https://github.com/Rafostar/clapper-vala-test))
d) Has overall more features (more expanded APIs then `GtkVideo` which is very simple/minimalistic)
e) Lets you customize/build your own overlay playback controls panel with buttons/features that best suit your app

[...]

PR #925, opened 2024-04-25
feat(Attachment.Box): stream attachments by default

rmader:
The bug preventing this before has long been fixed in Gst 1.22.1.

---

Closes https://github.com/GeopJr/Tuba/issues/924

### DIFF
--- a/src/Widgets/Attachment/Box.vala
+++ b/src/Widgets/Attachment/Box.vala
@@ -160,7 +160,7 @@ public class Tuba.Widgets.Attachment.Box : Adw.Bin {
 				false,
 				attachment_widgets[i].pic.alternative_text,
 				null,
-				false,
+				true,
 				is_main,
 				is_main == null
 			);


GeopJr:
Thanks!

Streaming *worked* but by using Files as a workaround. What I wanted to do was use inputstreams from libsoup. I pushed a commit that does that, but it won't work until it's fixed from GTK's side. They should be cacheable (this is just an initial implementation, I'd probably avoid creating yet another session for it and just use the images one)

If you try to play a video, you'll be met with a bunch of criticals and exit:

```
ERROR:../modules/media/gtkgstmediafile.c:290:gtk_gst_media_file_open: code should not be reached
Bail out! ERROR:../modules/media/gtkgstmediafile.c:290:gtk_gst_media_file_open: code should not be reached
Application exited
```

rmader:
Gave it a quick try with https://gitlab.gnome.org/rmader/gtk/-/commits/issue4062 (from https://gitlab.gnome.org/GNOME/gtk/-/issues/4062#note_2093829), but that'll need a bit more debugging.

That being said, with just the first commit things work quite well already for me (tested with flatpaks with gst 1.24 from https://github.com/rmader/Tuba/tree/flatpak on laptops and phones). I'm missing the background here, but why do you prefer downloading over that kind of streaming / why inputstreams needed?

GeopJr:
When adding PeerTube support, which mostly allows streaming, I discovered two major issues that forced me to disable it:
1. On long videos, sometimes the audio was not synced and almost every single one would stop updating its image after the \~30 minute mark while audio would still play - even after destroying the video widget
2. They weren't seekable, its length was unknown

Not really related to the media viewer, but when drafting animated emojis (which was also blocked), a MediaFile was needed and (due to that bug + an unfortunate issue with Vala) I had to also use GFiles. I hit the `file-max` limit pretty quick and killed tuba.

Looking into the future, even if the PeerTube issues didn't occur due to using GFiles, I believe having control over the requests is important:
1. ActivityPub software can deny requests if there's no user agent or one they don't like. GoToSocial for example does (or used to) deny all requests made without a user agent to their API
2. Supposedly someone makes a 'secure' AP server, that requires a token or a POST request to get assets, it wouldn't be possible with just GFiles (unless im wrong)

rmader:
I gave this a try on top of https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/7186 - *sometimes* it works and plays videos fine, but usually it crashes with:
```
libsoup:ERROR:../libsoup/cache/soup-cache-input-stream.c:154:soup_cache_input_stream_write_next_buffer: assertion failed: (priv->output_stream && !g_output_stream_is_closed (priv->output_stream))
Bail out! libsoup:ERROR:../libsoup/cache/soup-cache-input-stream.c:154:soup_cache_input_stream_write_next_buffer: assertion failed: (priv->output_stream && !g_output_stream_is_closed (priv->output_stream))
Aborted (core dumped)
```

or without any error at all. It sounds like some threading issue in the MR here to me, but we should probably figure it out first before merging the GTK MR.

rmader:
P.S.: thanks for the explanation, that makes a lot of sense!

GeopJr:
For the sake of not blocking it from the GTK side, I wrote a small reproducer that uses GFile but with inputstream https://github.com/GeopJr/Tuba/tree/experiment/reproducer-inputstream-video

the flatpak manifest uses your GTK branch

Unfortunately, I see it's just as unreliable, the video plays but the image is lagging behind. Seeking a bit can sometimes break it completely. Maybe streaming a local file is better for getting it merged on GTK

I experimented with libsoup threading but I faced other problems (plus libsoup is not entirely thread-safe) :upside_down_face:

I'll deal with these!

rmader:
> For the sake of not blocking it from the GTK side, I wrote a small reproducer that uses GFile but with inputstream 
> ...
> Unfortunately, I see it's just as unreliable, the video plays but the image is lagging behind. Seeking a bit can sometimes break it completely. Maybe streaming a local file is better for getting it merged on GTK

Nice, thanks a lot! So IIUC that means there's some issue in either libsoup, gtk or gstreamer not playing nicely yet - and that's not Tubas fault. So I'd say let's first try to fix that - we have the whole cycle before GTK 4.16 / Gnome 47 get released :)

rmader:
> I experimented with libsoup threading but I faced other problems (plus libsoup is not entirely thread-safe) 🙃

Just wanted to drop that I wonder if using the Gstreamer `souphttpsrc` would make sense - and if you tried that before?

GeopJr:
I haven't, but I'll experiment with it!

FWIW, I did start the initial experiment with Clapper on #931, and was pleasantly surprised with how well it worked

[Screencast from 2024-04-30 05-31-40.webm](https://github.com/GeopJr/Tuba/assets/18014039/bbfc5bd8-accf-4820-9504-9912e119bf46)

(ignoring the papercuts) streaming just worked, it was seekable, it was aware of its length etc. I do recognize however, that clapper being relatively new as a library (instead of an app) and requiring(?) some gstreamer patches for now, it cannot be the default (that's also why it falls back to GtkVideo if meson cannot find it). I'll dive a bit deeper at some point to see how they deal with streaming

rmader:
FYI., I'm just playing around with https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/7186 again as, while Clapper and the integration in Tuba is nice, I'd love to see vanilla GTK to provide the basic features here.

Issue #204, opened 2023-04-22
[Bug]: Should check for full network connectivity using NetworkManager before attempting anything


nekohayo:
### Describe the bug

If you launch Tuba while your computer is offline (i.e. not connected to wifi or anything), it will get stuck with the Home timeline loading with the spinner churning infinitely, even after the wifi has connected, even if you switch between tabs or sidebar categories in Tuba, when you go back to the "Home" area (or whatever you were trying to load while offline) it remains unable to retry loading it, it just keeps showing the blank slate and spinner.

So basically this is two things:
* Tuba should use NetworkManager to know if the user is fully online (not oflline, not just in a LAN/intranet, and not behind a captive portal). Here's some docs, but I don't know which part is the best way to go about it: https://networkmanager.dev/docs/api/latest/ - I am presuming you are not using NetworkManager because a quick github search didn't show mentions of "nm" (or something that sounded like Network Manager when searching for "network")
* Each component of the app should be network-failure resilient and reset its state after a while, or when switching away and back to it, or when using the future refresh/pull-down-refresh functionality, otherwise they might never really recover from that initial offline stage (in case of a bug) or from server-side failure.

### Steps To Reproduce

1. Launch Tuba while your computer is not connected to the Interwebs

### Instance Backend

Mastodon

### Operating System

Fedora 38

### Package

Flatpak

### Troubleshooting information

os: GNOME 44 (Flatpak runtime)
prefix: /app
flatpak: true
version: 0.2.0 (production)
gtk: 4.10.1 (4.10.1)
libadwaita: 1.3.1 (1.3.1)
libsoup: 3.4.0 (3.4.0)
libgtksourceview: 5.8.0 (5.8.0)

mlundblad:
Just beware that you might get bug reports from users using custom VPN scripts that don't play nicely with NM (or bypass it).

We had this in Maps with recurring bug reports. Added a --force-online CLI option as a workaround, but eventually bailed, and removed the network check (instead checking status when performing operations such as searching).

CaptainMorgan12:
i would not perform network checks within any apps, i think having and loadind old cached messages should be ok, either they are refreshed if network exists, or nothing new or additional is loaded if it doesnt.

nekohayo:
https://docs.gtk.org/gio/iface.NetworkMonitor.html might be an easier abstraction over NetworkManager, maybe.

GeopJr:
On #928 (draft), I ended up using NetworkMonitor.

[...]

PR #928, opened 2024-04-26
feat: network monitoring

GeopJr:
fix: #204 

Roadmap:

- [X] active account recovery (notifications, announcements, instance info, emojis)
- [X] websocket reconnections
- [x] timeline refreshes
- [x] check if you can access saved accounts at all (maybe don't even try until we are online?)
- [ ] finish offline window
- [ ] fix offline window closing => quitting (maybe an actual button?)
- [ ] avi recovery (maybe only for own accounts as timelines get refreshed?)
- [ ] recover any offline saved accounts
- [ ] tests tests and more tests

When the network changes from offline to online, it triggers a websocket and timeline refresh. That means we can skip recovering attachments and manually re-doing connections since they are getting refreshed.

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -12,6 +12,7 @@ namespace Tuba {
 	public static AccountStore accounts;
 	public static Network network;
 	public static Streams streams;
+	public static NetworkMonitor network_monitor;
 
 	//  public static EntityCache entity_cache;
 	//  public static ImageCache image_cache;
@@ -31,6 +32,7 @@ namespace Tuba {
 		public Dialogs.MainWindow? main_window { get; set; }
 		public Dialogs.NewAccount? add_account_window { get; set; }
 		public bool is_mobile { get; set; default=false; }
+		public bool is_online { get; set; default=true; }
 
 		public Locales app_locales { get; construct set; }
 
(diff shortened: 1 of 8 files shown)


GeopJr:
I think this insight is important https://github.com/GeopJr/Tuba/issues/204#issuecomment-1589488697

Maybe a runtime cli option to avoid checking altogether?

GeopJr:
Yeah no, network monitor is very inconsistent, we'll have to minimize the impact

GeopJr:
New issue: if the app starts offline, libsoup/glib/tls cannot recover and will "Temporary failure in name resolution" no matter what

GeopJr:
Added a gsettings option for disabling network monitoring

Not in preferences. If it needs to be in the GUI, then we need better naming because 'network monitoring' sounds malicious (see tracker-miner-fs)

GeopJr:
**Call for testing:**

Do you use a VPN or a proxy?
Is your internet connection very unstable?
Do you have an unusual setup?
Do you use something else than NetworkManager?

This PR could use your help!

- ## Flatpak

1. Download the PR artifact for your architecture, either from the GitHub CI or [nightly.link](https://nightly.link/GeopJr/Tuba/actions/runs/8899006367)
2. Unzip it
3. Run `flatpak install --user /path/to/dev.geopjr.Tuba.Devel.flatpak` (you can remove `--user` if all your runtimes are installed system wide to avoid excess downloads)
4. Test it
5. When you are done, remove it by running `flatpak remove tuba` and selecting the one you installed (usually the one that ends with `/master`)

- ## GNOME Builder

If you don't want to use the CI-built flatpaks, you can instead build it using GNOME Builder.

1. `git clone https://github.com/GeopJr/Tuba -b feat/network/monitoring --depth=1`
2. Open GNOME Builder and select the new folder
3. Click the play button at the headerbar (might require installing runtimes and will be greyed out until it's done)

- ## From source (advanced)

NetworkMonitor handles the flatpak differently than source, if you are comfortable with building Tuba, testing it outside of the sandbox would be useful. You can find build instructions in the README, but make sure you've switched to this branch.

# What to look for

If you experience disconnects, does Tuba recover gracefully after reconnecting? E.g. will you still be able to receive notifications and new posts?

If your setup is unusual, does Tuba correctly notice that you have or don't have internet connection?

Anything that should change in your opinion?

PR #947, opened 2024-05-17
feat(RichLabel): cleanup

GeopJr:
- Tuba will now open web+ap urls in-app (e.g. `web+ap://faraway.town/@GeopJr` will open my profile in-app instead of going through the process of asking your OS to open the default app for `web+ap` uris)
- Hashtag url resolving is now a lot more strict and valid (by using GLib.Uri)
- Host.open_uri was split into another method for opening GLib.Uris instead of strings
- Link resolving was optimized thanks to the above to use only one GLib.Uri instance instead of creating multiple for every single thing (checking the uri scheme to see if it's web+ap, tracking parameter splitting etc...)
- Fixed a critical in Utils/Tracking when attempting to iterate null query parameters
- Should_resolve conditions were optimized and became stricter

### DIFF
--- a/src/API/Notification.vala
+++ b/src/API/Notification.vala
@@ -123,11 +123,11 @@ public class Tuba.API.Notification : Entity, Widgetizable {
 	public override void open () {
 		switch (kind) {
 			case InstanceAccount.KIND_SEVERED_RELATIONSHIPS:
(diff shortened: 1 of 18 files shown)


GeopJr:
Terrible Utils.Host API names

Issue #932, opened 2024-05-01
[Bug]: Touchpad zoom gesture requires click in media viewer


eobet:
### Describe the bug

When opening an image on a laptop, zoom and scroll gestures on the touchpad don't work until you click once on the image. After you click it starts working. (Similarly, multi touch gestures on a touch screen I'd say are completely broken.)

### Steps To Reproduce

1. Open any image from any post.
2. Attempt to use the system standard pinch to zoom on a trackpad.

### Logs and/or Screenshots

```
<paste your logs here>
```

### Instance Backend

glitch-soc

### Operating System

Nobara (Fedora 39 w Gnome/Wayland)

### Package

Flatpak

### Troubleshooting information

os: GNOME 46 (Flatpak runtime)
prefix: /app
flatpak: true
version: 0.7.2 (production)
gtk: 4.14.3 (4.14.1)
libadwaita: 1.5.0 (1.5.0)
libsoup: 3.4.4 (3.4.4)
libgtksourceview: 5.12.0 (5.12.0)

### Additional Context

I'm using a Microsoft Surface Studio laptop with the special kernel extensions (already packaged with Noabra) which enables both the trackpad gestures and the multi touch tablet mode for the display).

GeopJr:
:/ I can't reproduce it here. Does your host/Nobara use KDE? is it Wayland or X11?

Ill see about setting up a VM to see if I can reproduce it there. From Tuba's side, it does everything it should be doing, setting the current focus to the media viewer so gestures can work ootb without extra clicks/manually changing the focus. I'll also see if setting the focus to one of the inner widgets, changes anything

> Similarly, multi touch gestures on a touch screen I'd say are completely broken.

By that, do you mean they also require an extra click/tap? "I'd say are completely broken" gives me the impression that they don't work at all

eobet:
Nobara uses KDE these days, but my install predates that, so I'm on a vanlilla Gnome variant with Wayland.

However, when I went today to try to replicate it, I failed. I've noticed with Blender that there's a difference in behavior if you start the app vs continuing to use it after you've waken the laptop from sleep. So maybe that's a Gnome/Wayland bug.

And yes, by "completely broken" I mean multi touch gestures via the touch screen does not work at all. That I confirmed today again, and I also tested with the built in Gnome Image Viewer and that was fully functional as expected.

PR #955, opened 2024-05-26
fix(MV): pass measure to scale revealer

GeopJr:
fix: #932 ?

MV is GtkWidget and should be measured by passing it to the scale revealer (AdwBin)


GeopJr:
Merging, anyone affected by the issues feel free to try this and re-open them if it didn't work. I couldn't reproduce either reliably before or after

PR #959, opened 2024-05-28
feat: Admin Dashboard

GeopJr:
This is going to take ages to finish because setting up a mastodon instance is extremely annoying.

Needs ~~\~16~~ \~4-5 pages

A lot to do but organized most of it so we can move forward to implementing the endpoints


GeopJr:
Finally! I'm free of this curse!

src/Dialogs/Admin/Report.vala
```diff
@@ -0,0 +1,376 @@
...
+
+		var status_group = new Adw.PreferencesGroup () {
+			title = _("Reported Posts")
+		};
+
+		report.statuses.foreach (status => {
+			try {
+				status.formal.filtered = null;
+				status.formal.spoiler_text = null;
+				Widgets.Status widget = (Widgets.Status) status.to_widget ();
```
ThisIsMissEm, line 272: In report UI's, do make sure to blur & greyscale media displayed to the user (ideally with toggles off for each independently), to avoid display harmful content directly to moderators

(mastodon's moderator tools currently only blur reusing the sensitive media mode, instead of also have greyscaling)

Looks like it took a heap of work to implement this; if there's APIs or things you'd need from the mastodon side for admin do let me know

GeopJr: Thanks! I'll move that to another issue so I can keep track of it.

> if there's APIs or things you'd need from the mastodon side for admin do let me know

I'm not an instance admin (also why I didn't think of or notice the blur and greyscale features) so I really don't know what instance admins would need from a native client either. As can be seen by the edit on the pr body, I went in this with the thought that I would be implementing *all* admin and mod sections, only to discover that most are not accessible by the api (everything under administration, invites...). I'm not sure if that's something an admin would want to do outside of mastodon-web (and tbh I'd rather Tuba wasn't responsible for any of these :))

Again, I really don't know if these are something admins would want, but some of the stuff I remember not being able to do through the API:
- Memoriam 
- Change user email
- Change user password
- Audit logs
- Notes
- Add/remove posts from a report

---

I found a bug while implementing updating reports, where if there are rule violations with selected rules, you cannot change category. Mastodon-web will show it as if it was changed, but api-wise it hasn't:

[Screencast from 2024-06-24 05-54-33.webm](https://github.com/GeopJr/Tuba/assets/18014039/16e1738c-a816-4fbf-a00a-69665422b3bd)

I was running a some-versions-behind instance when testing so I don't know if it has been fixed on master. From Tuba's side, I work around that by removing the rules before changing the category

ThisIsMissEm: This looks to maybe just be an awkward API without good validation for report category change, as that's the way Mastodon's own UI does it too: https://github.com/mastodon/mastodon/blob/4743657ba24e83c376e9f477fbf49114e6f09a57/app/javascript/mastodon/components/admin/ReportReasonSelector.jsx#L127

ThisIsMissEm: I think these might be most likely to be able to be added to the API:

- Audit logs
- Notes (assuming Account Notes and Report Notes)
- Add/remove posts from a report

GeopJr: Ah looks like the fix is still in master https://github.com/mastodon/mastodon/commit/9ce914cc897d2b1100cee8cb6d792a2283d877ec

Thanks for taking a look!

> I think these might be most likely to be able to be added to the API

Sounds good to me!

ThisIsMissEm: >  I'm not sure if that's something an admin would want to do outside of mastodon-web (and tbh I'd rather Tuba wasn't responsible for any of these :))

I'd agree tbh, as it is, I'm surprised you've implemented so much. Building good administrative and moderator UIs can be challenging because the needs are often very different to those of regular users.

PR #966, opened 2024-05-31
feat(PreviewCard): redesign

GeopJr:
After the discussion on https://gitlab.gnome.org/Teams/Circle/-/issues/193#note_2109127, I believe the vertical layout is more desirable nowadays. Not only does it match the other platforms but also websites usually have large and wide preview images.

Link with image

Screenshot from 2024-05-31 05-15-08

Link without image

Screenshot from 2024-05-31 05-13-08

Without an image seems a bit wasteful, we could make it use the horizontal layout then or remove the image part completely.

Additionally, I noticed that mastodon-web doesn't show preview cards if there are attachments which seems logical, so I made Tuba do it too.

We could also increase the description to 2+ lines since we now have the height freedom to do so

cc: @bertob

### DIFF
--- a/data/style.css
+++ b/data/style.css
@@ -209,7 +209,7 @@ headerbar.flat.no-title .title {
 	border-right: solid 1px @sidebar_border_color;
 }
 
-.ttl-profile-fields-box-container.odd > .ttl-profile-fields-box > flowboxchild:last-child > .ttl-profile-field { 
+.ttl-profile-fields-box-container.odd > .ttl-profile-fields-box > flowboxchild:last-child > .ttl-profile-field {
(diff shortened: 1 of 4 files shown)


bertob:
> Without an image seems a bit wasteful, we could make it use the horizontal layout then or remove the image part completely.

Using a horizontal layout with a fixed-width part on the left with the website icon makes sense to me.

> Additionally, I noticed that mastodon-web doesn't show preview cards if there are attachments which seems logical, so I made Tuba do it too.

+1

> We could also increase the description to 2+ lines since we now have the height freedom to do so

Makes sense, I'd try 2 or 3 lines and see how it looks.

PR #975, opened 2024-06-09
snap: update for core24

soumyaDghosh:


.github/workflows/build.yml
```diff
@@ -69,15 +68,18 @@ jobs:
         sed -i '/version: git/d' snapcraft.yaml
         yq -i 'del(.parts.cleanup)' snapcraft.yaml
       if: matrix.arch == 'aarch64'
-    - uses: diddlesnaps/snapcraft-multiarch-action@v1
-      if: matrix.arch == 'aarch64'
+    - uses: diddledani/snapcraft-multiarch-action@v1
```
GeopJr, line 71: The action is still under `diddlesnaps/snapcraft-multiarch-action`

.github/workflows/build.yml
```diff
@@ -69,15 +68,18 @@ jobs:
...
+      if: ${{ false }} # needs some changes in the action to incorporate core24
       id: snapcraft_docker
       with:
         architecture: arm64
     - uses: snapcore/action-build@v1
       id: snapcraft
       if: matrix.arch == 'x86_64'
-    - uses: actions/upload-artifact@v3
+      env:
+        SNAPCRAFT_ENABLE_EXPERIMENTAL_EXTENSIONS: 1
```
GeopJr, line 81: I can't ship it with this, can I?

build-aux/snap/snapcraft.yaml
```diff
@@ -1,9 +1,9 @@
 name: tuba
-base: core22
+base: core24
 version: git
 adopt-info: tuba
 
-grade: devel
+grade: stable
```
GeopJr, line 6: Is this needed?

.github/workflows/build.yml
soumyaDghosh: Ooo sorry, I mistakenly changed it.

.github/workflows/build.yml
soumyaDghosh: Yes you can, but for now only amd64 and just from github actions.

build-aux/snap/snapcraft.yaml
soumyaDghosh: devel snaps can't be pushed to stable to candidate channels

GeopJr:
Thanks for this PR!

FWIW, this action and snap are only for ci artifacts and development builds. The release snap config lives at https://github.com/GeopJr/tuba-snap/ (you don't have to open a PR there, I'll pick the changes from here!)

build-aux/snap/snapcraft.yaml
GeopJr: This is only for devel purposes though, the release config lives at https://github.com/GeopJr/tuba-snap/

soumyaDghosh:
> Thanks for this PR!
> 
> FWIW, this action and snap are only for ci artifacts and development builds. The release snap config lives at https://github.com/GeopJr/tuba-snap/ (you don't have to open a PR there, I'll pick the changes from here!)

You cannot directly build this in the `snapcraft.io` website builder. Because gnome extension still experimental.

build-aux/snap/snapcraft.yaml
soumyaDghosh: We can directly publish this snap to stable from here. As the snap is far behind right now.

GeopJr:
:/ disappointing

soumyaDghosh:
> :/ disappointing

Actually the extensions are getting a major overhaul on the graphics side, with new graphics drivers. So, it's taking this long.

https://github.com/canonical/snapcraft/pull/4744

build-aux/snap/snapcraft.yaml
soumyaDghosh: Added workflow to publish this to store.

soumyaDghosh:
You can get the login cradentials via this command,

```bash
snapcraft export-login --snaps=tuba \
  --acls package_access,package_release \
  SNAP_STORE_STABLE
```

Then set that in an environment variable named `STORE_LOGIN`

GeopJr:
I'll do so when I get on my snap environment! Thanks!

I have one more question though, if it compiles with -Ddevel=true, how do I make 'stable' builds from here? If I promote the edge one, it will be treated as devel and the action is not aware of repo releases. That's one of the reasons I split it into devel snap config and release snap config in a different repo.

soumyaDghosh:
> I'll do so when I get on my snap environment! Thanks!
> 
> I have one more question though, if it compiles with -Ddevel=true, how do I make 'stable' builds from here? If I promote the edge one, it will be treated as devel and the action is not aware of repo releases. That's one of the reasons I split it into devel snap config and release snap config in a different repo.

Okay, so even more conditioning! Ummmm... For now I removed the devel using sed, but I think better you could do is replicate this in the snap manifests repo. And remove things from here which you don't need anymore.

GeopJr:
Here's an idea: How about I unlink the tuba-snap repo from the snapcraft runners and we instead add an action there with (workflow_dispatch so I can trigger it manually) to build and publish snaps?

That way, the action on this repo can stay as devel and the action on tuba-snap will always be release only

GeopJr:
Merging, thanks!

I don't know if/when I'll set up action publishing on the tuba-snap repo.

GeopJr:
Thanks for letting me know!

Can I drop `SNAPCRAFT_ENABLE_EXPERIMENTAL_EXTENSIONS`?

soumyaDghosh:
No Sorry, my bad, not yet. It's not yet stable. I hope in the next release of snapcraft, it'll be. And yes, you can drop it once stable

Issue #983, opened 2024-06-10
0.8.0 Design changes review


GeopJr:
0.8 is approaching and so does the string (and UI) freeze. Let's summarize most of the design changes for one final review

- 'In-reply' indicator

Pro: you can quickly know if a post is part of a thread or a reply
Con: you don't know who it's a reply to. Getting the account name would require an extra request which I'd like to avoid as to not spam instances. We can however know if it's a reply to themselves

EDIT: nvm we can probably get the name from the mentions ^, I'll do so, so it becomes 'In Reply to ...'

![Screenshot from 2024-06-10 11-35-45](https://github.com/GeopJr/Tuba/assets/18014039/02e6887e-8d82-419c-ba1e-c29e5e9d220d)

- Profile notes

![Screenshot from 2024-06-10 11-34-06](https://github.com/GeopJr/Tuba/assets/18014039/63f0233c-0457-4b0e-8d39-098c29206cc8)

- Mini profile on right click

Pro: #495, you can quickly see if the person you are replying to is verified, their stats, their pronouns etc. The design is the same as the profile one but... smaller

![Screenshot from 2024-06-10 11-33-02](https://github.com/GeopJr/Tuba/assets/18014039/88c4dcea-58ca-437c-a4ec-0eb2a13df36b)

- Profile fields

(sizing has been fixed since the video below, now all fields have the same size)

[...]

PR #995, opened 2024-06-12
fix(notifications): move filters and clear to a dialog

GeopJr:
#983


rmader:
Thank you so much for this - being able to remove favorites is just awesome!

GeopJr:
I'm still a bit unsure of the way *new* notifications are/should be treated with filters, let me know if something feels off!

rmader:
Well, if you ask me I'll surely use the chance to give some quick feedback :P

1. I personally like to be able to completely mute certain things - i.e. if I mute likes/favorites, I want them to be gone and not e.g. still show up on the "Notifications" tab as number, what they currently still do.
2. IMO it would make sense to move "Push notifications" and "Filter notifications out" from the setting to the same menu. That would streamline things and actually also mirror what the web version does - which for me works pretty well.
3. IMO the "Clear notifications" button does not belong to the filter page. Did you consider putting it next to the filter button in the top bar of the main window - was it to noise? That's where I'd expect it to be, in any case. And not on the top-left, as it shortly was :)

That's just my non-designer but happy user perspective, thanks for all your work!

GeopJr:
> I personally like to be able to completely mute certain things - i.e. if I mute likes/favorites, I want them to be gone and not e.g. still show up on the "Notifications" tab as number, what they currently still do.

Agreed, though I don't think I can completely get rid of it. The way marking notifications as read works, is by setting a 'marker' to the last read notification. So, while I can avoid increasing the count when Tuba received a filtered notification type, it won't be able to do it if the marker increased while offline. E.g. You make a post and close Tuba. All notification have been read. While offline, you get 10 favourites. When you re-open Tuba the marker will be less than the last notification id, so now there are 10 unread notifications you can't see because they are filtered...

Also the fact that 'Push Notifications' and 'Filtered Notifications' are not synced also complicated things. If you can receive a favorite push notification but at the same time have it filtered in notifications, what happens to the notification badge? You *received* a notification so it should increase but also you *filter it out* so it should :woozy_face: 

I think Push and Filtered notifications should be synced for starters and then deal with the notification badge, hopefully nobody will miss them being separate settings

> IMO it would make sense to move "Push notifications" and "Filter notifications out" from the setting to the same menu. That would streamline things and actually also mirror what the web version does - which for me works pretty well.

I brought it up on #983

>> I'm still not sure about this being a dialog. Should all notifications settings move to it from settings? Probably not. But at the same time, there are different notification-related settings in two places now. On a second thought, these are specific to this view only so it makes sense 🤷

I'm conflicted between:
1. Having all settings in the settings dialog makes sense
2. Having notification settings in the notifications view also makes sense

I'll probably get notes on this after it gets shipped!

> IMO the "Clear notifications" button does not belong to the filter page. Did you consider putting it next to the filter button in the top bar of the main window - was it to noise? That's where I'd expect it to be, in any case. And not on the top-left, as it shortly was :)

That's how it was originally but it doesn't fit in narrow mode :laughing:

Sidebar toggle button, clear notification button, title, filter button, search button (visible in narrow mode, might remove it based on Tobias' notes), window controls (especially annoying when all 3 are enabled)

image

Issue #1042, opened 2024-06-26
[Bug]: The margin of the popover when clicking on an avatar is odd


Kekun:
### Describe the bug

In messages, when clicking on the avatar of its author there's a popover that gives information about this author's account. This popover has some odd looking margin between its border and its content, it breaks the look of the scrolled window, its shadows, its scrollbar is oddly in the middle, etc.

### Steps To Reproduce

1. Go to some message
2. Click the author's avatar to reveal the popover

### Logs and/or Screenshots

![Capture d’écran du 2024-06-26 10-19-47](https://github.com/GeopJr/Tuba/assets/2536339/868a23b0-f841-4ab3-a4d3-5740a7f29abe)
![Capture d’écran du 2024-06-26 10-19-20](https://github.com/GeopJr/Tuba/assets/2536339/9f8f4f1b-7f0b-47b5-ad0f-de37ffbef62e)

### Instance Backend

Mastodon

### Operating System

Fedora Linux 40.20240617.0 (Silverblue)

### Package

Flatpak

### Troubleshooting information

os: GNOME 46 (Flatpak runtime)
prefix: /app
flatpak: true
version: 0.8.1 (production)
gtk: 4.14.4 (4.14.4)
libadwaita: 1.5.1 (1.5.1)
libsoup: 3.4.4 (3.4.4)
libgtksourceview: 5.12.0 (5.12.0)

GeopJr:
The padding is the default style for all popovers, could you point me on the exact issues / what should change and how?

For example, in your screenshot, I see these borders looking out of place 
![image](https://github.com/GeopJr/Tuba/assets/18014039/0f76cf8b-78a5-4d9c-aa2d-b0c7753a8f6f)

And these don't really match 
![image](https://github.com/GeopJr/Tuba/assets/18014039/c50ca7e7-8f53-4212-90e5-f7cc40ed82b3)

Border radius should probably be dropped or set it to a more appropriate value, like the account switcher popover 
![image](https://github.com/GeopJr/Tuba/assets/18014039/bc969c79-e4ec-4e7d-9520-dff413c4b759)

Kekun:
Yeah I tried to quickly hack a thing in the inspector when writing this report, but I didn't manage to do a thing because I couldn't get the popover to stay there while I was browsing the inspector. I'll have to build the app and all, which is more involved and will take me more time, so that'll be when I'll be able to find the time for it.

GeopJr:
[...]

PR #1043, opened 2024-06-26
fix(Avatar): drop the popover's padding (#1042)

Kekun:
### Before

Capture d’écran du 2024-06-26 19-27-57
Capture d’écran du 2024-06-26 19-28-19

### After

Capture d’écran du 2024-06-26 19-28-45
Capture d’écran du 2024-06-26 19-28-58

### DIFF
--- a/data/style.css
+++ b/data/style.css
@@ -98,10 +98,6 @@ headerbar.flat.no-title .title {
 	background: @dialog_bg_color;
 }
 
-.large .header-image {
-	border-radius: 12px 12px 0 0;
-}
-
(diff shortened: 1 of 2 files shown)


data/style.css
```diff
@@ -98,10 +98,6 @@ headerbar.flat.no-title .title {
 	background: @dialog_bg_color;
 }
 
-.large .header-image {
```
GeopJr, line 101: This was mostly meant for the focus ring, but it looks like it broke at some point, I'll fix it in a separate PR, so I'll revert this

PR #1044, opened 2024-06-27
feat(mv): Use "Clapper.MediaItem.cached" constructor

Rafostar:
Simplify code by storing only URIs and their cache file locations as strings, then use newly added [Clapper.MediaItem.cached](https://rafostar.github.io/clapper/doc/clapper/ctor.MediaItem.new_cached.html) constructor to recreate an item.

NOTE: This PR targets `feat/mv/enable-clapper-caching` branch.


GeopJr:
Looks good to me! Thank you so much!

Issue #1025, opened 2024-06-22
[Bug]: Crash after clicking on embedded media into Mastodon's toot


FYHenry:
#### From Tuba

In rare cases click on embedded media into Mastodon toot can force Tuba to crash.
The crash is triggered just after zooming the media.
The playing overlay appears too before Tuba crashes.

Tuba shows a similar log like the issue #1006 (*offload on*).

#### From Firefox

From the Web browser the media can be read without error.
The media format is MP3 or PNG.

### Steps To Reproduce

1. Run `flatpak run Tuba dev.geopjr.Tuba`,
2. Search URL [https://ludosphere.fr/@PolGM/112660922165799640](https://ludosphere.fr/@PolGM/112660922165799640) from the Search tab,
3. Open the *toot*,
4. Click one of the four embedded medias.

#### Debug log (with `G_MESSAGES_DEBUG`)

Substitution words : `USERNAME`, `HOSTNAME` and `CDNPATH`.

```
(dev.geopjr.Tuba:2): Tuba-DEBUG: 22:59:13.217: Application.vala:247: os: GNOME 46 (Flatpak runtime)
(dev.geopjr.Tuba:2): Tuba-DEBUG: 22:59:13.217: Application.vala:247: prefix: /app
(dev.geopjr.Tuba:2): Tuba-DEBUG: 22:59:13.217: Application.vala:247: flatpak: true
(dev.geopjr.Tuba:2): Tuba-DEBUG: 22:59:13.217: Application.vala:247: version: 0.8.0 (production)
(dev.geopjr.Tuba:2): Tuba-DEBUG: 22:59:13.217: Application.vala:247: gtk: 4.14.4 (4.14.4)
(dev.geopjr.Tuba:2): Tuba-DEBUG: 22:59:13.217: Application.vala:247: libadwaita: 1.5.1 (1.5.1)
(dev.geopjr.Tuba:2): Tuba-DEBUG: 22:59:13.217: Application.vala:247: libsoup: 3.4.4 (3.4.4)
(dev.geopjr.Tuba:2): Tuba-DEBUG: 22:59:13.217: Application.vala:247: libgtksourceview: 5.12.0 (5.12.0)
(dev.geopjr.Tuba:2): Tuba-DEBUG: 22:59:13.220: SecretAccountStore.vala:9: Using libsecret v0.21.4
[...]

Issue #1048, opened 2024-06-29
[Bug]: Opening videos more than once crashes the app (failed to flush Wayland connection)


atirut-w:
### Describe the bug

This applies to *any* video.

### Steps To Reproduce

1. Open a video from a post
2. Go back and open the video again or any other video
3. The app crashes.

### Logs and/or Screenshots

You can see the crash happening in this repro video:

https://github.com/GeopJr/Tuba/assets/25323231/2ce9242c-acc0-4640-aad5-fb1f82a1f8b8

```
(dev.geopjr.Tuba:2): Tuba-DEBUG: 17:58:24.346: Host.vala:58: Downloading file: https://static.toot.community/cache/media_attachments/files/112/699/180/094/107/125/original/36f5517ae6e4afe6.mp4…
(dev.geopjr.Tuba:2): Tuba-DEBUG: 17:58:24.346: Host.vala:88:    OK: File already exists
(dev.geopjr.Tuba:2): Tuba-DEBUG: 17:58:27.384: MediaViewer.vala:238: Destroying MediaViewer.Item
(dev.geopjr.Tuba:2): Tuba-DEBUG: 17:58:28.455: Host.vala:58: Downloading file: https://static.toot.community/cache/media_attachments/files/112/699/180/094/107/125/original/36f5517ae6e4afe6.mp4…
(dev.geopjr.Tuba:2): Tuba-DEBUG: 17:58:28.455: Host.vala:88:    OK: File already exists

(dev.geopjr.Tuba:2): GStreamer-GL-CRITICAL **: 17:58:28.736: Failed to flush Wayland connection


(dev.geopjr.Tuba:2): GStreamer-GL-CRITICAL **: 17:58:28.736: Failed to flush Wayland connection


(dev.geopjr.Tuba:2): GStreamer-GL-CRITICAL **: 17:58:28.736: Failed to flush Wayland connection

Gdk-Message: 17:58:28.737: Error flushing display: Invalid argument
```

### Instance Backend

Mastodon

### Operating System

Fedora Linux 40 (Workstation Edition)

### Package

Flatpak

### Troubleshooting information

os: GNOME 46 (Flatpak runtime)
prefix: /app
flatpak: true
version: 0.8.1 (production)
gtk: 4.14.4 (4.14.4)
libadwaita: 1.5.1 (1.5.1)
libsoup: 3.4.4 (3.4.4)
libgtksourceview: 5.12.0 (5.12.0)

[...]

PR #1049, opened 2024-06-29
fix(mv): set use-graphics-offload to false

GeopJr:
fix: #1025 #1048

Graphics offload is awesome but the random segfaults under *some* hardware are just so out of my control that I'd rather disable it for now.

Initially, the issue was the loading stack, but even after removing that, it still crashes on other people's hardware (it fixed it for me).

It's just not worth it. I can't even create a reproducer easily.


rmader:
This is likely caused by https://gitlab.gnome.org/GNOME/gtk/-/issues/6824 and also affects other apps. Hopefully offload can be turned on again once things have been further stabilized.

Issue #435, opened 2023-08-06
[Request]: Looping animated images treated as such (instead of treated as videos)


nekohayo:
### Describe the request

If you look at https://mastodon.social/@jmsdnns/110843691825850858 through Tuba, it will behave as a 1-second long video that immediately stops after starting.

Animated GIFs, WebP, JPEG XL (now supported by the latest WebKitGTK for GNOME 45), AVIF, should be treated as such. 

Maybe autoplaying while:
* on a mobile device/formfactor (because there is no such thing as hovering, as there is no mouse...)
* on a desktop device that is not in powersaver energy mode nor disable/reduce-animations mode; hover-to-play otherwise.

And ideally / if needed for power savings and performance (i.e. if it's not done automagically by the toolkit) stop the playback when scrolled off-screen.

### Implementation Details

- [ ] This should be an option in settings.
- [ ] This should be only available to some fediverse backends. (Include which ones on the above field).
- [X] This is client-only (and shouldn't sync with the instance).
- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

GeopJr:
Took a closer look for #682 

The main blockers are https://gitlab.gnome.org/GNOME/gtk/-/issues/4062 and the whole performance issue:

1. All GIFs are actually GIFVs (that's why they are treated as videos). Due to the linked bug (which is actually a Gstreamer issue (https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/724)), all videos on Tuba get downloaded rather than streamed. Streaming GIFVs should be more appropriate
2. (I've gone into multiple rants about this already but) since the posts are not getting recycled, having them be downloaded or streamed in full just so we can play them on hover would increase memory usage by a lot

PR #1060, opened 2024-07-08
feat(mv): Few small clapper code enhancements

Rafostar:
Few small changes/enhancements for Clapper MediaViewer code:
* Enable loop for GIF videos
* Complement MPRIS args
* Do not show controls for GIFV

### DIFF
--- a/src/Views/MediaViewer.vala
+++ b/src/Views/MediaViewer.vala
@@ -965,6 +965,7 @@ public class Tuba.Views.MediaViewer : Gtk.Widget, Gtk.Buildable, Adw.Swipeable {
 
 			if (media_type == Tuba.Attachment.MediaType.GIFV) {
 				#if CLAPPER
+					video.player.queue.progression_mode = Clapper.QueueProgressionMode.REPEAT_ITEM;
 					video.player.autoplay = true;
 				#else
 					video.loop = true;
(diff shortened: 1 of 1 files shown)


GeopJr:
LGTM, thanks!

Rafostar:
My main aim here was to add loop for GIFV which was missing. I also changed for GIFV to not show bottom controls panel as browsers do not do that and I think this is right thing to do, but if you disagree let me know and I can drop this commit.

GeopJr:
Oh no, it's great, you actually fixed #435 (partially)! (Removing Gtk.Video's controls would be too annoying so I never did it)

That being said, that change might be changed in the future if users complain because Mastodon has video limitations.

On Mastodon you can only post either one video or 4(+) attachments, so if you want to post videos and other attachments, users convert them to gifs. Those gifs are sometimes too long and technical and could benefit from video controls, e.g. the first attachment here: https://mastodon.social/@haeckerfelix/112656595629880944 ... But, overall treating them as gifs is the best option!

Rafostar:
On the side note: The looping should be "seamless" (without hiccups) after combined with #1010. Then video will be streamed at the beginning followed by looping by using cached file exclusively. But these two PRs are not depending on each other as that part is handled transparently inside Clapper 0.8 so they can go in in any order.

PR #1064, opened 2024-07-15
Revert "fix(mv): set use-graphics-offload to false (#1049)"

rmader:
The previously observed crashes were most likely all cases of a wrong assert in the GTK offload code that affected multiple apps, fixed in https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/7407

Thus turn on offloading again.

This reverts commit 050f49c0e086eac5d2cac6d630cc63677e6b2f2f.

---

At least https://github.com/GeopJr/Tuba/issues/1025 ran into the assert fixed in the MR above. I've previously been running into the crash as well and tested the commit in this MR for a couple of days without any more problems, so I guess it things should be safe to reenable.

### DIFF
--- a/data/dev.geopjr.Tuba.gschema.xml
+++ b/data/dev.geopjr.Tuba.gschema.xml
@@ -104,7 +104,7 @@
 			<default>''</default>
 		</key>
     <key name="use-graphics-offload" type="b">
-			<default>false</default>
+			<default>true</default>
 		</key>
 
 		<key name="window-w" type="i">


GeopJr:
Thanks for taking a look!

I think I'll leave it as is for now and instead add a meson check to just get rid of the setting altogether if gtk >= 4.16 (plus re-add all the removed stuff from #1008 and #1007, which I would also like to test on 4.16)

PR #1070, opened 2024-07-18
fix(translation): Add context to tracking param string 

bragefuglseth:
Just stumbled upon the tracking param stripper in action – thanks for having that! – and noticed that my translation of the string was wrong as I had translated "Stripped" as an adjective rather than as a verb. This PR adds a comment clarifying what to do for translators.


GeopJr:
LGTM, thanks!

PR #1075, opened 2024-07-26
fix(mv): clapper: GIFV handling fixes

Rafostar:
Some fixes mainly related to handling multiple GIFV in a single post. More info in individual commits descriptions.

---

Ideally, Tuba could be more efficient by creating a single video instance, putting multiple GIFV in Clapper [Queue](https://rafostar.github.io/clapper/doc/clapper/class.Queue.html) and just switching between them (maybe even with overlaid pre-made [NextItemButton](https://rafostar.github.io/clapper/doc/clapper-gtk/class.NextItemButton.html) / [PreviousItemButton](https://rafostar.github.io/clapper/doc/clapper-gtk/class.PreviousItemButton.html)) instead of creating multiple video instances and switching widgets. Unfortunately, this would be probably too much/hard to implement while retaining support for GtkVideo which is the default backend within a single source code file, so for now a few changes working around that to remain compatible :smile: .

### DIFF
--- a/src/Views/MediaViewer.vala
+++ b/src/Views/MediaViewer.vala
@@ -944,14 +944,6 @@ public class Tuba.Views.MediaViewer : Gtk.Widget, Gtk.Buildable, Adw.Swipeable {
 				var video = new ClapperGtk.Video () {
 					auto_inhibit = true
 				};
-				#if CLAPPER_MPRIS
-					var mpris = new Clapper.Mpris (
-						"org.mpris.MediaPlayer2.Tuba",
-						Build.NAME,
-						Build.DOMAIN
-					);
-					video.player.add_feature (mpris);
-				#endif
 				video.player.audio_filter = Gst.ElementFactory.make ("scaletempo", null);
 			#else
 				var video = new Gtk.Video () {
@@ -974,6 +966,13 @@ public class Tuba.Views.MediaViewer : Gtk.Widget, Gtk.Buildable, Adw.Swipeable {
 				#endif
 			} else {
 				#if CLAPPER
+					#if CLAPPER_MPRIS
+						video.player.add_feature (new Clapper.Mpris (
+							"org.mpris.MediaPlayer2.Tuba",
+							Build.NAME,
+							Build.DOMAIN
+						));
+					#endif
 					video.add_fading_overlay (new ClapperGtk.SimpleControls () {
 						valign = Gtk.Align.END,
 						fullscreenable = false
--- a/src/Views/MediaViewer.vala
+++ b/src/Views/MediaViewer.vala
@@ -958,10 +958,8 @@ public class Tuba.Views.MediaViewer : Gtk.Widget, Gtk.Buildable, Adw.Swipeable {
 			if (media_type == Tuba.Attachment.MediaType.GIFV) {
 				#if CLAPPER
 					video.player.queue.progression_mode = Clapper.QueueProgressionMode.REPEAT_ITEM;
-					video.player.autoplay = true;
 				#else
 					video.loop = true;
-					video.autoplay = true;
 				#endif
 			} else {
 				#if CLAPPER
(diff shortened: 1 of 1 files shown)


GeopJr:
LGTM, thanks!

I want to do some deeper testing with some extreme cases and I'll merge it afterwards

> Ideally, Tuba could be more efficient by creating a single video instance, putting multiple GIFV in Clapper [Queue](https://rafostar.github.io/clapper/doc/clapper/class.Queue.html) and just switching between them (maybe even with overlaid pre-made [NextItemButton](https://rafostar.github.io/clapper/doc/clapper-gtk/class.NextItemButton.html) / [PreviousItemButton](https://rafostar.github.io/clapper/doc/clapper-gtk/class.PreviousItemButton.html)) instead of creating multiple video instances and switching widgets. Unfortunately, this would be probably too much/hard to implement while retaining support for GtkVideo which is the default backend within a single source code file, so for now a few changes working around that to remain compatible 😄 .

Unfortunately there's a lot at play :(, for example, even the attachment position matters and if there's: GIFV, GIFV, IMAGE, GIFV, then it would need at least 2 Clapper instances. Position matters sometimes, like in the case of documents where each attachment is a different page etc.

Maybe we could use a queue if all the attachments are videos? (but that will complicate things a lot, so let's just leave as is for the near future :sweat_smile:)

Rafostar:
> so let's just leave as is for the near future

Absolutely. Doing it like it is done currently was my intention, hence this PR.

> GIFV, GIFV, IMAGE, GIFV

In ideal scenario, there would be exactly 1 video player instance (since Tuba shows only 1 video/GIFV at a time) and media viewer is something that can show both videos and images (e.g. a GTK Stack with video and picture viewers that switches between these 2 child widgets when needed). If it could hide itself instead of destruction when unneeded/done that would be even better. There is a lot of work going on (between GTK and GStreamer) to prepare for video playback (not to mention decoders and separate memory pools of each pipeline, etc.). Afterwards, most of it does not have to be done again when video widget is recycled. This is true for both GtkVideo and Clapper that use GStreamer internally, so both would benefit.

I just wanted to mention that. This is just a friendly TIP, for how this can be optimized (how can but its not a must, just optimization). I do not want nor plan to meddle with someone else's app too much. Just trying to give some friendly pointers/help and nothing more :smile: .

src/Views/MediaViewer.vala
```diff
@@ -103,10 +103,12 @@ public class Tuba.Views.MediaViewer : Gtk.Widget, Gtk.Buildable, Adw.Swipeable {
...
 					#if CLAPPER
+						Clapper.Player player = ((ClapperGtk.Video) child_widget).player;
+
 						if (value) {
-							((ClapperGtk.Video) child_widget).player.play ();
-						} else {
-							((ClapperGtk.Video) child_widget).player.pause ();
+							player.play ();
+						} else if (player.state > Clapper.PlayerState.STOPPED) {
+							player.pause ();
```
Rafostar, line 111: I am not sure if desired outcome when switching between videos/GIFV is to pause previous one and later resume (like its currently done) or to just stop and play from beginning when selected again?

GeopJr:
> I just wanted to mention that. This is just a friendly TIP, for how this can be optimized (how can but its not a must, just optimization). I do not want nor plan to meddle with someone else's app too much. Just trying to give some friendly pointers/help and nothing more 😄 .

Thank you for all the info! I value your opinion and knowledge on the matter and really appreciate it :) Feel free to share your thoughts whenever you feel like it!

src/Views/MediaViewer.vala
GeopJr: I haven't put much thought into it, but the only thing I can think of in favor of pausing, is that someone might be watching a long video or gif and accidentally swipe to the next attachment, losing the video progress

Rafostar:
Seems to work fine (at least for me :smile:) and does what its supposed to do (no MPRIS for GIFs and no buffering/download for remaining videos/gifs when only one is selected), so I am unmarking this from being a draft.

GeopJr:
LGTM, thanks!

PR #1085, opened 2024-08-12
feat(search): bold matches

GeopJr:
Let's make search query keywords bold!

Screenshot from 2024-08-13 00-39-23

But it's not ready:

Screenshot from 2024-08-13 00-57-12

the above problem is that we need to re-calculate the indexes based on the new text, `this is a test and another test` => `this is a <b>test</b> and another test` at this point the length of the string has changed, and we need to get the next `test`'s index from the new string

Problem being, this can be quite time consuming, might be better to just use regex instead.

Also from the first screenshot we can see that join**mastodon** is marked. 'mastodon' is inside a string and while it showed up, it probably only did because of the account name and not the link. We should probably only take into account words that start with the keyword

`mastodonCats` :heavy_check_mark: 
`meowMastodon` :x:

and lastly, words that are <= 3 characters should probably only count as whole

"foo **to** bar" :heavy_check_mark: 
"**to**mato" :x:

This is a draft, but the foundation has been laid and only the details are left

### DIFF
--- a/src/API/Account.vala
+++ b/src/API/Account.vala
@@ -1,5 +1,6 @@
-public class Tuba.API.Account : Entity, Widgetizable {
+public class Tuba.API.Account : Entity, Widgetizable, SearchResult {
 	public API.Relationship? tuba_rs { get; set; default=null; }
+	public string? tuba_search_query { get; set; default = null; }
 
 	public string id { get; set; }
 	public string username { get; set; }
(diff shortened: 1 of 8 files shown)


GeopJr:
> Problem being, this can be quite time consuming, might be better to just use regex instead.

and so I did

The regex gets compiled when searching with the 'optimize' flag and then gets passed along to all results

GeopJr:
Merging, if anyone tests this let me know of any shortcomings :)

PR #1086, opened 2024-08-21
fix(alt-text-dialog): flat style margin top

GeopJr:
For some reason, the scrolledwindow having a margin top makes the undershoot appear with a padding

https://gitlab.gnome.org/Teams/Circle/-/issues/193#note_2202009


alice-mkh:
Yeah, undershoot is a part of the scrolled window, not of the toolbar view or the header bar. So it's whenever your scrolled window ends

GeopJr:
Thanks for the info!

PR #1087, opened 2024-08-21
feat(MainWindow): replace navigation stack when going back to start

GeopJr:
Can't remember if there was a reason we popped everything instead of replacing. The current popped signal handler is only used for widget focus wizardy, maybe we should just call on_popped in go_back_to_start?

This removes the back animation
https://gitlab.gnome.org/Teams/Circle/-/issues/193#note_2202009


alice-mkh:
Yup, that's precisely the kind of thing replace() is for ^^

PR #1099, opened 2024-09-02
feat(mv): audio visualizer

GeopJr:
I learnt more about gstreamer than I thought I would going into this...

Okay so, audio visualizer. Originally used cairo but converted it to snapshots because getting paintables into an imagesurface is more complicated than expected

If there's a thumbnail, it uses it + its blurhash to get the average color and themes the visualizer accordingly. If the color is too dark, it lightens it up.

https://github.com/user-attachments/assets/da4d77c8-96ce-4923-a5f6-490f5846a340

If there's no thumbnail, it will attempt to use the author's avi. I decided to go with accent color for the visualizer. With adw 1.6 we can get the custom ones too.

https://github.com/user-attachments/assets/5415a5ab-9d5c-4bc9-9aa9-b794a2d5b980

TODO:

- [x] hide controls on non-movement
- [x] leak test
- [x] test mv on multi-format media
- [x] flatpak + meson for gstreamer
- [x] adw 1.6 accent on non-thumbnail
- [x] fix any deprecation warnings if possible
- [x] keyboard shortcuts
- [x] volume slider adjustment feels too large?
- [x] insensitive until ready?

### DIFF
--- a/.github/workflows/msys2.yml
+++ b/.github/workflows/msys2.yml
@@ -20,7 +20,7 @@ jobs:
         with:
           msystem: mingw64
           update: true
-          install: git make zip wget mingw-w64-x86_64-meson mingw-w64-x86_64-gcc mingw-w64-x86_64-vala mingw-w64-x86_64-libsoup3 mingw-w64-x86_64-libxml2 mingw-w64-x86_64-gtksourceview5 mingw-w64-x86_64-webp-pixbuf-loader mingw-w64-x86_64-libadwaita mingw-w64-x86_64-libgee mingw-w64-x86_64-json-glib mingw-w64-x86_64-libsecret mingw-w64-x86_64-desktop-file-utils mingw-w64-x86_64-imagemagick mingw-w64-x86_64-icu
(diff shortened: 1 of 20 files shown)


GeopJr:
Wondering if I should introduce gstreamer as a required dependency or not

GeopJr:
cc: @bragefuglseth if you have any design input on the visualizer

Fwiw this is mastodon web and twitter


https://github.com/user-attachments/assets/604a837c-5a78-46f9-a759-0f15828506eb


https://github.com/user-attachments/assets/addb02ae-3ae6-497b-98f8-69d1093d99c8

bragefuglseth:
This looks cool! Great initiative.

One thing that stands out to me is that the up/downscaling of the circle feels a little exaggerated; I'd maybe decrease its max size to about half of what it is currently.

To make the color blend better with the pitch black background, maybe the circle could also be made partially transparent (0.5)?

bragefuglseth:
The color change is good.

Just built and tested the PR, and I found that something closer to the original height makes a lot more sense when you're actually hearing the sound. The silent preview clips threw me off :D

I set MAX_CIRCLE_HEIGHT to 1024, and that seems to work well.

Issue #173, opened 2023-04-07
[Request]: Ability to specify the focal point of each image being attached when posting a status / toot


nekohayo:
### Describe the request

Since you can't know what the heck Mastodon is going to do to your images, you often have to click into the image to set the focal point while you are creating (or [editing](https://github.com/GeopJr/Tuba/issues/172)) a status. This controls, among other things, the portion of the image that gets shown in the cropped/partial (non-expanded) thumbnails in the timeline.

It would be awesome if Tuba would support setting and displaying this (I noticed that Tuba does not render the panned position of the thumbnail correctly for posts made by Mastodon's web UI with the focal point set).

Mastodon documentation on the matter:

* [Guidelines and best practices - focal points](https://docs.joinmastodon.org/api/guidelines/#focal-points)
* [Media API methods - Form data parameters](https://docs.joinmastodon.org/methods/media/#form-data-parameters)

I have no idea if other protocols support this.

### Implementation Details

- [ ] This should be an option in settings.
- [ ] This should be only available to some fediverse backends. (Include which ones on the above field).
- [ ] This is client-only (and shouldn't sync with the instance).
- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

PR #1101, opened 2024-09-06
feat(composer): focus picker

GeopJr:
## The focus picker

Mastodon allows you to pick the thumbnail focus of your attachments. It's based on https://github.com/jonom/jquery-focuspoint#1-calculate-your-images-focus-point

The focus point is communicated in the form of coordinates:

image

(0,0) being the center of the image, (1,1) the top right corner and (-1,-1) the bottom left one

The FocusPicker widget is responsible for displaying an image and moving the picker but also for calculating this mess.

API coordinates <=> width, height coordinates

The width,height coordinates have to be recalculated when the window gets resized (etc) so it stays in place as the image also gets resized

Mastodon also does this thing where you can focus outside of the image - even though the actual coordinates are image size based, as seen in its preview:

image

Enough talk, demo:

https://github.com/user-attachments/assets/08f29073-f2a7-4c6b-8039-35359e53494c

Notice how it stays in position between resizing, plus it stays in the limits of the image

TODO:

- [x] Finish the dialog
- [x] Accessibility (probably a scale for keyboard navigation)
- [x] Sends the coordinates to the API
- [x] Re-use the coordinates on edit (will probably have to make the pos_x/y props public or a helper function)
- [x] Select an actual good looking picker and make it circular 
- [x] Hide picker overflow on bottom and right (shouldnt be visible outside the picture)

fix: #173


GeopJr:
I think it's almost done, not really sure about the picker anymore but :shrug: 

This needs fixing too:

image


https://github.com/user-attachments/assets/61d1f998-6f41-438c-bd2b-7ca53f7385f8

GeopJr:
@rmader is there something broken with the focus picker?

GeopJr:
No worries about the noise! I assume it was the AudioVisualizer's adw 1.6 accent colors, I guess-worked how the binding *should* look like on release but I was wrong (actually I'm right and the bindings are incomplete but I'm not going to argue (`accent_color_rgba` should have been a property, instead I have to call the getter `get_accent_color_rgba()`))

PR #1113, opened 2024-09-14
feat(explore): link timelines

GeopJr:
The upcoming mastodon versions adds link timelines. That required quite a bit of refactoring as we now need a nested button to access them.

So preview cards were split to preview card (status sub-widget), preview card internal and preview card explore.

TODO: figure out when link timelines are available. To that I think the new api_versions of instance v2 could be helpful.


GeopJr:
locked to api_versions mastodon >=1

that is not 100% accurate, as link timelines work on tech.lgbt (glitch-soc) even though the api_versions commit hasnt made it in yet.

PR #1118, opened 2024-09-20
feat(ci): Post GNOME 47 runtime updates

Rafostar:
With GNOME 47 runtime, latest GStreamer version is included so we do not have to build anything else for Clapper anymore. Remove GStreamer from CI building and move Clapper to main manifest where all the other sources are.

GNOME 47 runtime is based on Freedesktop runtime 24.08, so we should be using a ffmpeg-full version made for this parent runtime.


Rafostar:
Huh. Seems that flatpak-builder in git already has 47 (and its in README), but stable release from git actions marketplace does not. We can either wait with this MR until its updated in marketplace or just drop last commit and continue with 43 for now...

GeopJr:
Awesome, thanks!

> just drop last commit and continue with 43 for now...

Yeah, just drop it. The action version has little to do with the actual runtimes. The main branch already depends on GNOME 47's libraries!

Rafostar:
OK. Lets just update the builder action while at it.

Rafostar:
Tested with Builder and works great. :+1: 
I also fixed MPRIS permission that we missed last time while adding multiple instances support.

GeopJr:
Thank you so much! I'll probably include it in the main flatpak in the next release :)

Rafostar:
> I'll probably include it in the main flatpak in the next release

Awesome, but... there still isn't a Clapper stable that would have downloads cache ability unfortunately :disappointed: (it helps a lot for looping "GIF-videos").

GeopJr:
I don't have a schedule for the next release, but I can switch the flatpak to Clapper whenever cache makes it to stable without having to make a new Tuba release (since we've already implemented everything on Tuba's side)!

Issue #1122, opened 2024-09-22
[Bug]: Images going fullscreen and back animate to stretched fullscreen rather than to target best fit size


YaLTeR:
### Describe the bug

When clicking on an image, it crossfades into a fullscreen media viewer, but the image itself crossfades into a fullscreen stretch. Instead, it should crossfade into its target best fit location.

### Steps To Reproduce

1. Click on some image whose aspect ratio significantly mismatches the Tuba window aspect ratio.

### Logs and/or Screenshots

https://github.com/user-attachments/assets/2eb7a714-f174-4f38-b7ba-e7155011eee5

You can see the problem clearly on freeze frames like:

![mpv-shot0001](https://github.com/user-attachments/assets/132d6d16-4bf2-4079-9f15-c94fbbcb5c91)

### Instance Backend

Mastodon

### Operating System

Fedora 40 Silverblue

### Package

Flatpak

### Troubleshooting information

os: GNOME 46 (Flatpak runtime)
prefix: /app
flatpak: true
version: 0.8.4 (production)
gtk: 4.14.5 (4.14.4)
libadwaita: 1.5.4 (1.5.3)
libsoup: 3.4.4 (3.4.4)
libgtksourceview: 5.12.1 (5.12.1)

GeopJr:
My initial thought is that it's not worth it (unless I'm misunderstanding what the desired outcome is).

As it stands, the ScaleRevealer (the 'animation'), works by snapshotting the source widget and the target widget and crossfading between them.

The main issue is that there are way too many moving parts to have an 'expand to target size' animation work reliably and seamlessly.

From Tuba's side, there are options to letterbox attachment thumbnails and not enlarge media in the media viewer. That means that when viewed in the media viewer, images can actually be smaller than that the thumbnail.

From fedi backends' side, thumbnail sizes might differ from the full-size attachment, video thumbnails are often square and do not match the first frame of the video. Audio files usually don't have thumbnails and the media viewer will display an audio visualizer anyway.

I can play around with it and see if maybe snapshotting the child widgets or maybe changing when the crossfade starts will have a nicer effect, but I'd rather avoid doing more complex snapshot manipulation

YaLTeR:
Hm, I understand that there are difficulties here, but maybe preserving the thumbnail aspect ratio during the crossfade would already make it better?

GeopJr:
[...]

PR #1125, opened 2024-09-22
feat(ScaleRevealer): finish crossfade early

GeopJr:
fix: #1122 ?

Previously the transition would crossfade between the source widget (attachment) and the target one (mediaviewer) relatively to the transition progress. That creates a seamless transition between them. However as shown on #1122, it also creates an unintended effect around the middle point.

For context, the media viewer's black background's opacity is also relative to the animation.

This PR changes it so as long as the animation progress is over 20%, it won't crossfade. This still looks a bit off, maybe just to me because I've seen it in slow mode.

Another idea would be to make the media viewer's background's opacity get darker sooner so the distorted snapshot of the attachment is not that visible?

whichever the path forward might be, this PR can be optimized further so it's a draft. (we can skip snapshoting the source on over 20% since the fade effect is done).

### DIFF
--- a/src/Widgets/ScaleRevealer.vala
+++ b/src/Widgets/ScaleRevealer.vala
@@ -138,12 +138,9 @@ public class Tuba.Widgets.ScaleRevealer : Adw.Bin {
 			warning ("The source widget texture is None, using child snapshot as fallback");
 			this.snapshot_child (this.child, snapshot);
 		} else {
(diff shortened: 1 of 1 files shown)


GeopJr:
Made it ease in-out so it hits more progress points, made progress limit 0.3 and optimized snapshoting to skip the whole cross fading process as we only need to snapshot the child

PR #1133, opened 2024-09-25
feat(Profile): avi and header alt text

GeopJr:
GTS added these recently on their rc

TODO:
- test on gts


GeopJr:
confirmed working on gts

PR #1142, opened 2024-09-28
feat(preferences): remove options

GeopJr:
Following options are unpopular or a default value is clearly the preferred one. This PR will remove them.

If you use any of these, I'm open to leaving it in gsettings without a UI toggle so **speak now or forever hold your peace**

| Removed Option | Default | Comments |
| :---: | :---: | :---: |
| color scheme | following system theme | - |
| show preview card | true | - |
| scale emoji on hover | true | - |
| media viewer expand pictures | true | - |
| use blurhash | true | - |

### DIFF
--- a/data/dev.geopjr.Tuba.gschema.xml
+++ b/data/dev.geopjr.Tuba.gschema.xml
@@ -28,9 +28,6 @@
 		<key name="active-account" type="s">
 			<default>''</default>
 		</key>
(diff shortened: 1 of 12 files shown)


LukaszH77:
I use "media viewer expand pictures"

GeopJr:
Would you mind if it was gsettings only or do you think it should stay as a switch in the GUI?

LukaszH77:
I'm fine with using gsettings to turn that off.

PR #1143, opened 2024-09-29
feat(about): Show other apps

bragefuglseth:
In GNOME 48, libadwaita will ship with a new, cool feature that allows developers to show off their other projects in the about window of the app. This is an initial implementation of that in Tuba.

Currently it requires a custom manifest to build due to some ongoing difficulties with abstract Vala interfaces / the GNOME master runtime, but that can be dropped before merging this when the GNOME 48 runtime becomes available.

Whether you want this or not is of course also up to you, there can be a lot of reasons not to :slightly_smiling_face: 

Skjermbilde fra 2024-09-29 20-30-43


GeopJr:
Thanks!

I'll revisit this in... 6 months! :laughing:

bragefuglseth:
haha, just performing some early field testing :)

GeopJr:
LGTM, thanks!

image

Issue #880, opened 2024-03-30
[Request]: Support less prominent hashtags


bertob:
### Describe the request

The Mastodon website now displays hashtags at the bottom of the post in a less prominent way (outside the regular content area and below media). I'm not sure what their exact heuristic for this is, but it might just be literally filtering out the last line of the post if it's all hashtags?

It'd be cool to do something similar in Tuba as well, it makes posts a lot more legible.

![image](https://github.com/GeopJr/Tuba/assets/1908896/db61c715-aa06-4280-9f09-922969f8edbc)

![image](https://github.com/GeopJr/Tuba/assets/1908896/861b0a7a-1313-4cd0-8d7e-14242cca5a76)

### Implementation Details

- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

GeopJr:
Should the design be the same as #555?
And, should we show the first 3 and require a button press for the rest like Mastodon does?

bertob:
Yeah, +1 on both

PR #1149, opened 2024-10-07
feat: TagExtractor

GeopJr:
fix: #880 

I tried out different ways to do this and honestly this is probably the best.

- Extracting while parsing the HTML

The way libxml2 traverses through a doc is unpredictable (aka, when it reaches an anchor, you can't exactly get the content, you have to go through it and usually the `#` character and the hashtag name are in different elements). That makes it impossible to detect that an anchor is a hashtag quickly and before parsing and appending it. Same thing for detecting if the last paragraph only has hashtags.

The structure also depends on the backend.

- Extracting them afterwards by parsing the pango markup

While we now have a predictable structure, the same issue as above is still present. It becomes too difficult to reliably detect if the last paragraph is only made of hashtag anchors.

- Current solution

Regex. (Anchors are predictable as it's being ran on the converted HTML)

https://github.com/user-attachments/assets/6bb4f161-a3f0-4757-89d5-cb5b6c672e18

### DIFF
--- a/data/style.css
+++ b/data/style.css
@@ -645,6 +645,10 @@ video > overlay > revealer > controls, .audio-controls {
 	border-radius: 9999px;
 }
 
+.hashtag-bar-hashtag {
+	padding: 4px 11px;
+}
+
 .preview_card:not(.explore), .preview_card:not(.explore) image:not(.attachment-overlay-icon) {
 	background-color: alpha(currentColor, 0.0625);
 }
--- a/data/ui/widgets/status.ui
+++ b/data/ui/widgets/status.ui
@@ -321,6 +321,7 @@
                                       <object class="TubaWidgetsMarkupView" id="content">
                                         <property name="visible">True</property>
                                         <property name="hexpand">False</property>
+                                        <property name="extract-last-tags">True</property>
                                       </object>
                                     </child>
                                   </object>
--- a/src/Utils/TagExtractor.vala
+++ b/src/Utils/TagExtractor.vala
@@ -1,0 +1,54 @@
+public class Tuba.TagExtractor {
+	public struct Tag {
+		public string tag;
+		public string link;
+	}
+
+	public struct Result {
+		public string? input_without_tags;
+		public Tag[]? extracted_tags;
+	}
+
+	static GLib.Regex pango_uri_regex;
+	static construct {
+		try {
+			pango_uri_regex = new GLib.Regex ("<a href='(?<href>http[^']+)'>#(?<tag>[^<]+)<\\/a>", GLib.RegexCompileFlags.OPTIMIZE | GLib.RegexCompileFlags.CASELESS);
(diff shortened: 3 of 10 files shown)


GeopJr:
> it'd be good to avoid the different widths

(you are aware of flowbox' width situation, where all columns have the same width which is the biggest item's. So the widths will be different when there are multiple rows)

> but maybe we could do something like the Mastodon web UI and only show 4 by default, and the rest in a popover or something

The video is a bit quick but the implementation is the same as mastodon-web. It shows the first 3 and then I click the 'Show 5 More' button which causes them all to appear. No opinion on popover or even dialog

Here's some different styles:

<details><summary>Smaller font</summary>

Screenshot From 2024-10-08 11-37-16

</details>

<details><summary>Less padding + Smaller font</summary>

Screenshot From 2024-10-08 11-37-10

</details>

<details><summary>Normal font weight</summary>

Screenshot From 2024-10-08 11-36-32

</details>

<details><summary>Normal font weight + Less padding</summary>

Screenshot From 2024-10-08 11-36-23

</details>

<details><summary>Less padding</summary>

Screenshot From 2024-10-08 11-36-12

</details>

bragefuglseth:
> (you are aware of flowbox' width situation, where column items have the same width which is the biggest item's. So the widths will be different when there are multiple rows)

FYI: [this just landed in libadwaita’s main branch](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.WrapBox.html)

GeopJr:
Should I port it over? I have many use cases for WrapBox that are currently using FlowBox here

bragefuglseth:
Sure, it would be good to get some in-the-wild usage of the widget early on! And it seems like an obvious choice for this particular thing.

GeopJr:
After spending some time porting it, I'm not sure if I want to continue.

We could use LabelWithWidgets in the meantime however:


https://github.com/user-attachments/assets/eb821f0e-0d2d-4bca-857b-a53b5967fbf4

since it's not really made just for custom emojis, but for any widget in general

bragefuglseth:
That works for now (with more space between the elements, haha)

GeopJr:
Performance is terrible, I'll stick to flowbox for now :shrug: :(

bertob:
> Less padding + Smaller font

This one looks good to me!

> After spending some time porting it, I'm not sure if I want to continue.

If you encounter any issues with WrapBox, now would be a great time to provide feedback :)

GeopJr:
> If you encounter any issues with WrapBox, now would be a great time to provide feedback :)

Nothing to do with WrapBox itself! Porting it to Vala is quite the challenge due to some incompatibilities between C and Vala and ripping it out of libadwaita as is requires copying over a lot of other things, so it's probably best to wait for 1.7

bertob:
Ah yeah, that makes sense :+1:

Issue #197, opened 2023-04-18
[Bug]: Underline missing on multiline links


bertob:
### Describe the bug

Looks like some multi-line links are missing the underlining on the first line (built in Builder):

![image](https://user-images.githubusercontent.com/1908896/232802175-197cd9f6-accf-40e1-b97e-6229d8850837.png)

### Steps To Reproduce

Open the app, find a post with a multi-line link.

### Instance Backend

Mastodon

### Operating System

Fedora 38 Beta

### Package

Compiled manually

### Troubleshooting information

os: GNOME 44 (Flatpak runtime)
prefix: /app
flatpak: true
version: feat/status/title-redesign-fcc3ecd (development)
gtk: 4.10.1 (4.10.1)
libadwaita: 1.3.1 (1.3.1)
libsoup: 3.4.0 (3.4.0)
libgtksourceview: 5.8.0 (5.8.0)

GeopJr:
It has to do with the font-size, increasing it to 15px seems to get rid of it but more investigation might be needed

![image](https://user-images.githubusercontent.com/18014039/232805811-6b5d5a77-a57e-4a45-a259-387dd3a45619.png)

PR #1150, opened 2024-10-07
feat: do not force GL by default

GeopJr:
fix: #197 

When ngl was the default, some users noticed a significant performance regression. So we set the default to be GL (while respecting `GSK_RENDERER`).

On 47, Vulkan became the default renderer and the performance should be much better than ngl.

Both ngl and Vulkan fix many artifacts (like curcular avatars becoming square while scrolling sometimes) and the #197 issue from what I gathered.


GeopJr:
@LukaszH77 if I remember correctly, you noticed a performance regression with `ngl` back when it became the GTK default. Could you try vulkan (the new default) whenever you get the chance to and tell me if you notice anything?

(You don't have to use this branch, you can do it on your currently installed build by running it with `GSK_RENDERER=vulkan <command to run tuba>`)

Issue #914, opened 2024-04-18
[Bug]: A post causes a UI overflow on Tuba App


kop316:
### Describe the bug

Hello!

This post seems to cause an overflow of the UI of the Tuba App: https://akkoma.meows.gay/objects/b0dcc9c5-ece9-49eb-a83a-7c2b8b60c1bd

Screenshot:
![Screenshot_2024-04-17_20-40-09](https://github.com/GeopJr/Tuba/assets/16343055/c09bce5f-12af-4377-bc8f-45d7d533ae54)

### Steps To Reproduce

1. Follow that account
2. 2. Have it show up on your Home feed.

### Logs and/or Screenshots

```
<paste your logs here>
```

### Instance Backend

Mastodon

### Operating System

Debian Trixie

### Package

OS repositories

GeopJr:
That's not it (or at least it's doesn't reproduce it here)

In general, the issue is with GtkLabel + ellipsize + wrap measuring the widget size incorrectly. GtkLabels are beyond broken under multiple containers, see this one for another example https://github.com/GeopJr/Tuba/issues/756#issuecomment-1912794500

It has been reported again #719 and again #632 and again #626 and many other times, but unless I get the post (or rather label) actually causing it, reproducibly, I can't really do anything, including creating a minimal reproduction so gtk fixes it or a workaround from our side

Finding which one is actually causing it is also somewhat difficult as you have to play around with the inspector until you figure it out, which is likely to crash a few times. I don't expect anyone to go through that but I'm still throwing it out there just in case

GeopJr:
Though, from the looks of it you are also using an older version of Tuba, it might have been fixed on 0.7 :shrug:

kop316:
Ahh, thanks for that. You are correct, that was the wrong post. I think I found the correct post that causes it: https://fosstodon.org/deck/@yassie_j@labyrinth.zone/112287918084026585

kop316:
It looks like since this is a duplicate, it doesn't make sense to keep open.

GeopJr:
Thanks for figuring it out!

I'll re-open this as they are mostly case-by-case. I'll see about writing a workaround!

GeopJr:
Okay so, GtkLabel reports a larger min width that natural. This can be fixed by removing the overflow workaround but then we need to fix that. So for now, I'll say blocked by https://gitlab.gnome.org/GNOME/gtk/-/issues/6637 and see how it goes

GeopJr:
[...]

PR #1172, opened 2024-10-21
fix(lww): label measuring when pango wrap is word_char

GeopJr:
fix: #914 

TODO: maybe set it to word only when inside markupview?


GeopJr:
Haven't noticed any regressions so far.

Issue #1173, opened 2024-10-22
[Bug]: "Following" panel in user profile is not updated when (un)following someone


heldderarbeit:
### Describe the bug

The user profile shows the number of followed accounts, that number is not updated when following or unfollowing someone while the application is running.

### Steps To Reproduce

1. Click on the hamburger menu - Open profile
2. remember the number in the middle panel (xx Following)
3. Follow or unfollow someone inside Tuba
4. Go back to the profile and look at the following panel

Expectation: the number should be updated
Result: the number was not updated

After restarting, the number is shown correctly.

### Logs and/or Screenshots

```
<paste your logs here>
```

### Instance Backend

Mastodon

### Operating System

Arch

### Package

Flatpak

### Troubleshooting information

os: GNOME 46 (Flatpak runtime)
prefix: /app
flatpak: true
version: 0.8.4 (production)
gtk: 4.14.5 (4.14.4)
libadwaita: 1.5.4 (1.5.3)
libsoup: 3.4.4 (3.4.4)
libgtksourceview: 5.12.1 (5.12.1)

GeopJr:
Nice catch, it should be fixed by #1174, thanks!

PR #1174, opened 2024-10-22
feat(profile): refresh user when appropriate

GeopJr:
this required more effort than expected

ProfileCover should now be update-able outside of construct. Also, refreshing a profile view will now also refresh the profile cover.

To combat unnecessarily updating the cover, own profiles (the active profile) gets updated on load, this is because an older version is probably saved in secrets. Other profiles will only be updated when manually refreshing as they should be recent.

fix: #1173


GeopJr:
I need to do a leak test and generally check if anything is broken

PR #1192, opened 2024-11-06
Let BlurHash to be transliterated for non Latin scripts.

danialbehzadi:
in many non-Latin scripts e.g. Persian, words like `BlurHash` should be transliterated or they can not be read, understood or identified at all.

### DIFF
--- a/data/ui/dialogs/preferences.ui
+++ b/data/ui/dialogs/preferences.ui
@@ -76,7 +76,7 @@
 						</child>
 						<child>
 							<object class="AdwSwitchRow" id="use_blurhash">
-								<property name="title" translatable="no">BlurHash</property>
+								<property name="title" translatable="yes">BlurHash</property>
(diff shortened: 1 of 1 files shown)


GeopJr:
IMO it's more of a 'trademark'. It's like translating 'Apple', the company, to 'apple', the fruit.

BlurHash is a product/tool (https://blurha.sh/) and not a generic spec or description.

The row subtitle explains what a blurhash is which is probably more useful to users than translating BlurHash as blur + hash.

danialbehzadi:
I think the distinction is that it's transliteration, not translation. For example, 'Apple' is branded as اپل in Persian, which is pronounced similarly to the original, but has no relation to the fruit 'apple' which is written as سیب. Similarly, transliterating 'BlurHash' would help users recognize the brand/product name, even if the individual words don't have a direct meaning in Persian.

GeopJr:
I'm still on the side that it might cause confusion but giving the option to translators wouldn't hurt I guess

(I'll add a comment recommending leaving it as is later)

Issue #1191, opened 2024-11-06
[Request]: Support Mastodon's native grouped notifications.


danialbehzadi:
### Describe the request

Mastodon 4.3 introduced a new grouped notifications on `/api/v2/notifications*`

Here is the documentation: https://docs.joinmastodon.org/methods/grouped_notifications/

### Implementation Details

- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

Issue #176, opened 2023-04-09
[Request]: compact / concise notifications


dilinger:
### Describe the request

Currently every like or boost or follow gets the entire post listed in a notification. Having a popular toot easily overwhelms notifications. Elk.zone solves this by combining likes/boosts for the same toot, and also doesn't display images for ones own toots, like so:

![image](https://user-images.githubusercontent.com/11581730/230750184-ff642ff3-9258-4564-aa97-7aa28788a5ff.png)

Quite a difference compared to Tuba:

![image](https://user-images.githubusercontent.com/11581730/230750227-70ca96b2-c5d3-424e-85a1-23527dccc033.png)

I would suggest two things:

1) Don't display images when a user's own toot is boosted/liked. I wrote the toot, I know the content, I don't need my notifications filled with the same image(s) over and over.  I'd suggest still continuing to display images for other circumstances (eg, someone replies with an image); just not for toots authored by the main user.

2) Combine multiple likes and boots for the same toot into one notification entry. It doesn't have to be fancy (like, I wouldn't try and combine them out-of-order or anything; if 3 people boost Toot A, then someone boosts Toot B, and then another person boosts Toot A, there's no need to show Toot A with 4 boots in the same entry; just showing a notification timeline of "3 boosts for A, 1 boost for B, and 1 boost for A" would be a huge improvement). It could be as simple as something like the following, which I just mocked up (somewhat awkwardly) in gimp:

![Untitled](https://user-images.githubusercontent.com/11581730/230750841-4275e2a7-7419-48a7-9b6f-48a1bff136b9.png)

I'm unsure if this should be an option in settings or just the default. I dislike Elk's super-concise notifications (I do want to see the names of people who boosted/liked, not just tiny profile pic images). But I think if the same information were retained while also removing redundant stuff (so implementing the suggestion to combine notifications, while not implementing the suggestion to not display images), there would be no need for a settings option.

### Implementation Details

- [ ] This should be an option in settings.
- [ ] This should be only available to some fediverse backends. (Include which ones on the above field).
- [X] This is client-only (and shouldn't sync with the instance).
- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

nekohayo:
[...]

PR #1193, opened 2024-11-06
feat(notifications): groups

GeopJr:
fix: #1191 #176

Very far from done.


GeopJr:
it's getting shape but I still need:

1. do actual widgets, this 'fit group notifications in normal ones' is pretty bad
2. organize the version keeping mechs better
3. make sure the upgrade is seamless

GeopJr:
I quite hate it because it's messy when you want to support both v1 and v2, but whatever at this point...

should be done
Screenshot From 2025-05-18 00-35-38

GeopJr:
Some other stuff:

- I never noticed it before but it seems that the notifications websocket never emits updates and deletions. We need to deal with them however for privacy reasons (aka if someone mentions u and updates or deletes the post, youll still have it there...). To solve that we are forwarding events from the home websocket
- to_widget was marked as throwing an Oopsie. This was because it's an abstract method and we needed the empty one to always throw. But that's a programmer error if they haven't implemented it. The right solution is to assert instead, so we don't have to try and catch everywhere especially since we know it will never error.

PR #1196, opened 2024-11-09
Prevent reloading timeline after network outage

farao:
Since there is already an automatic reconnect mechanism of the websocket in Connection, there is no need to remove the subscription and empty and reload the timeline after a network outage.


src/Services/Network/Streamable.vala
```diff
@@ -60,15 +60,12 @@ public abstract interface Tuba.Streamable : Object {
...
 			update_stream ();
-		} else {
-			unsubscribe ();
 		}
 	}
 
 	protected void update_stream () {
 		// debug (get_subscriber_name ()+": UPDATED to "+subscribed.to_string ());
 
-		unsubscribe ();
```
GeopJr, line 71: `update_stream`, as the name implies, is used to update a stream e.g. when you change account

GeopJr:
Thanks for the PR!

I added my comments on the issue

PR #1205, opened 2024-11-17
feat(ProfileCover): birthday

GeopJr:
Account creation birthday style :shrug: 

also finally converted almost everything to css vars from gtk vars


LukaszH77:
Great idea, but rn it looks kinda weird on fresh profiles w/o any fields filled in. I think in this case it should be the width of the entire row. 

Zrzut ekranu z 2024-11-18 20-27-40

GeopJr:
Nice catch, I'll get on it!

Issue #142, opened 2023-03-31
[Request]: Scheduling / scheduled / delayed toots (and editable queue)


nekohayo:
### Describe the request

Mostly self-explanatory ;) for many, many, many reasons, I would like to be able to draft posts and schedule them for automatic publication at a certain date and time in the future.

The Mastodon protocol [has the ability to handle it](https://github.com/mastodon/mastodon/pull/9706) (but their web UI [doesn't expose the feature](https://github.com/mastodon/mastodon/issues/10088) yet). I don't know if the protocol allows managing/editing the [queue](https://github.com/mastodon/mastodon/issues/15219) of scheduled toots, though.

### Implementation Details

- [ ] This should be an option in settings.
- [ ] This should be only available to some fediverse backends. (Include which ones on the above field).
- [ ] This is client-only (and shouldn't sync with the instance).
- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

HallaSurvivor:
I would love, in particular, to be able to queue or schedule my boosts so that I don't feel like I'm spamming my followers whenever I'm online.

PR #1209, opened 2024-11-19
feat: scheduled posts

GeopJr:
fix: #142 

TODO:

- Scheduled posts view needs work
- Re-schedule


LukaszH77:
Nice! I'm not sure if it's something I would use on a regular basis, but it's a great tool to have. 

I tried it already and it works! \o/
Found one weird bug. I created the test post at 22:52 and in the Scheduled Post view it said I did it at 22:11. It got published at the right time, so that's good.

BTW: is the minimum delay of 6 (I think) minutes is something from Mastodon's defaults, or is it set in Tuba?

GeopJr:
> I'm not sure if it's something I would use on a regular basis, but it's a great tool to have.

It's a step towards draft posts too, which I'm a bit iffy about... Mastodon Android seems to use scheduled posts for drafts, but scheduled for 5024 (3000 years from now!), it's smart in the way that
1) makes sure that the post will be accepted by the server
2) drafts are synced between devices
3) media attachments will be uploaded

But... It also means that the instance has your drafts and scheduled posts are being abused for something they are not made for. 🤷 I guess I'll do it the same way, hopefully tuba and mastodon no longer exist in the year 5000

> I created the test post at 22:52 and in the Scheduled Post view it said I did it at 22:11.

Interesting, in that view, the timestamps are returned by the server, I wonder if they do something weird with the timezone, I'll check tomorrow!

> BTW: is the minimum delay of 6 (I think) minutes something from Mastodon's defaults, or is it set in Tuba?

The API claims it's 5, I did `>` instead of `>=` so it became 6 but Mastodon Android has a 10 minute minimum :/

GeopJr:
> I wonder if they do something weird with the timezone, I'll check tomorrow!

... welp, can't believe nobody noticed but formatting the date to `%H:%m` is Hour:MONTH (hence 11 => November). The correct one is `%M`. I partially blame valadoc because they are both lowercase https://valadoc.org/glib-2.0/GLib.DateTime.format.html

TODO: fix all translations...

Issue #1214, opened 2024-11-22
[Request]: Make scheduled posts editable.


LukaszH77:
### Describe the request

Could it be possible to be edit scheduled posts? 

### Implementation Details

- [X] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

GeopJr:
Technically the API only allows us to re-schedule them, but, I can probably hack it together so it deleted the one you are 'editing' and creates a new one when you save

PR #1218, opened 2024-11-23
feat(ScheduledStatuses): edit

GeopJr:
fix: #1214 

As per my comment on that issue, the API does not allow us to directly edit them but hacking around that, we can delete and reschedule them.

It gets a bit messy with schedule timing however. Since the min is 5 mins, we *cannot* edit scheduled posts that are about to be posted, users first need to reschedule them to a future date and the edit button will appear.

It also leaves a period of error: If a post is to be posted in 6 minutes and the user clicks edit, if they spend 6 minutes in the composer, it will be posted. If they spend 2 minutes in the composer, it won't be able to edit it as the new schedule time will only be 4 minutes in the future.


GeopJr:
Maybe we should increase the edit min to like an hour.

PR #1224, opened 2024-11-27
Document GtkSourceView 5 requirement in README.md

smyru:


GeopJr:
Thanks!

I went ahead and updated the whole section as well. Dropped the 'required version' column as it's always latest / I can't keep up with library versions, we always aim for the latest.

And I included `icu` and the optional deps

smyru:
I know it's hard to track all dependencies, but this information was very helpful for package maintainers.

Issue #1258, opened 2024-12-11
[Bug]: Home key used in the editor scrolls the timeline in the main window


LukaszH77:
### Describe the bug

Pressing `Home` while editing a post scrolls to the top of the timeline in the main window, instead of moving the cursor to the beginning of the text. The same with `Page Up` and `Page Down` but these are not as important as `Home`, I think.

### Steps To Reproduce

1. Open the post editor
2. Type some text
3. Press `Home` to go to the beginning
4. The text cursor doesn't move, instead the timeline in the main window scrolls up.

### Logs and/or Screenshots

```
<paste your logs here>
```

### Instance Backend

Mastodon

### Operating System

openSUSE

### Package

Flatpak

### Troubleshooting information

os: GNOME 47 (Flatpak runtime)
prefix: /app
flatpak: true
version: main-d23027ff (development)
gtk: 4.16.7 (4.16.7)
libadwaita: 1.6.2 (1.6.2)
libsoup: 3.6.1 (3.6.1)
libgtksourceview: 5.14.2 (5.14.2)
libspelling: true
Clapper: false
GStreamer: GStreamer 1.24.9 (1.9.24.0)

LukaszH77:
`Home` key doesn't seem to work for search either.

PR #1251, opened 2024-12-09
Add Home shortcut for scroll to top of page.

toolstack:
Add support for the Home key to take you to the top of the scroll window.

Issue #1055, opened 2024-07-03
[Bug]: I can open many dialogs many times


Kekun:
### Describe the bug

There is a short time window before the modal dialog appears where you can still click the Reply button, so if you double-click it you end up with two modal dialogs one on top of the other.

Similarly, if I spam Ctrl+, I can generate as many Preferences dialog as I want, and with F1 I can create as many About dialogs as I want. I suspect the problem happens pretty much everywhere in the app.

I think Tuba should better track its state, especially regarding opened dialogs. The simple way to do this is to store that dialog's object in a field of the object that creates it, to ignore any subsequent creation requests if there already is a dialog, and to clear that field when the dialog is destroyed. That way you can only have a single instance of each dialog.

### Steps To Reproduce

1. Double-click Reply on a message

1. Press F1 several times

1. Press Ctrl+, several times

### Logs and/or Screenshots

[Capture vidéo du 2024-07-03 10-06-17.webm](https://github.com/GeopJr/Tuba/assets/2536339/48def1c1-045a-4eae-aabb-5bb02b7512d4)

### Instance Backend

Mastodon

### Operating System

Fedora Linux 40.20240702.0 (Silverblue)

### Package

Flatpak

### Troubleshooting information

os: GNOME 46 (Flatpak runtime)
prefix: /app
flatpak: true
version: 0.8.1 (production)
gtk: 4.14.4 (4.14.4)
libadwaita: 1.5.1 (1.5.1)
libsoup: 3.4.4 (3.4.4)
libgtksourceview: 5.12.0 (5.12.0)

GeopJr:
They are all the aftermath of moving to AdwDialogs from modals. Modals, being different windows, you also get the 'shortcut focus' so the app's shortcuts wouldnt work on them/while they are visible/while they are focused. But the new Dialogs are shown in the main window itself.

Not sure if that's something that should be dealt on libadwaita's side (I don't have the time to track it down rn), but it's common among all AdwDialog apps

[Screencast from 2024-07-03 11-17-09.webm](https://github.com/GeopJr/Tuba/assets/18014039/8a610b21-9307-4939-b4a9-be804f782105)

The reply one, I can easily deal from our side I guess, but the others should probably be dealt across all apps. A quick fix from apps' side would be checking if `Adw.Window#get_dialogs` is empty

Kekun:
[...]

PR #1259, opened 2024-12-12
revert: "feat: scroll to top on Home (#1251)"

GeopJr:
This reverts commit 7f8c8d333cc96d18024e5b9b1afb5b0c7836e864.

revert: #1251
fix: #1258 

Global action + AdwDialogs = bad. I quickly tried setting it as a view action or a shortcut, same results. No time to debug this rn we are too close to a release

### DIFF
--- a/data/gtk/help-overlay.ui
+++ b/data/gtk/help-overlay.ui
@@ -75,12 +75,6 @@
 								<property name="action-name">app.scroll-page-down</property>
 							</object>
 						</child>
-						<child>
-							<object class="GtkShortcutsShortcut">
-								<property name="title" translatable="yes">Scroll to Top</property>
-								<property name="action-name">app.scroll-page-to-top</property>
-							</object>
-						</child>
 					</object>
 				</child>
 				<child>
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -78,7 +78,6 @@ namespace Tuba {
 			{ "back-home", back_home_activated },
 			{ "scroll-page-down", scroll_view_page_down },
 			{ "scroll-page-up", scroll_view_page_up },
-			{ "scroll-page-to-top", scroll_view_page_to_top },
 			{ "goto-notifications", goto_notifications },
 			{ "open-status-url", open_status_url, "s" },
 			{ "answer-follow-request", answer_follow_request, "(ssb)" },
@@ -328,7 +327,6 @@ namespace Tuba {
 			set_accels_for_action ("app.back-home", {"<Alt>Home"});
 			set_accels_for_action ("app.scroll-page-down", {"Page_Down"});
 			set_accels_for_action ("app.scroll-page-up", {"Page_Up"});
-			set_accels_for_action ("app.scroll-page-to-top", {"Home"});
 			add_action_entries (APP_ENTRIES, this);
 
 			if (settings.monitor_network)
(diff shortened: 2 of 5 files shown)


toolstack:
@GeopJr if you let me know what the issue is I can take a look.  The home routines are basically the same as the pgup/pgdn code.

toolstack:
Also confirmed that pgup/dn are broken in dialogs as well.

GeopJr:
We are a bit cornered with this one. Moving the action to the view *should* be the appropriate solution but it's not, because the view needs to be focused for them to work but IIRC people complained about the pgup/dn buttons not working when they are focused on the sidebar for example.

<kbd>Home</kbd> specifically is a bit more critical because it's used in text fields to go back to the start *and* moving at the top can be far more annoying than an unwanted pgup/dn.

Technically this wouldn't be an issue prior to Adw.Dialogs, since the `app` actions wouldn't be trigger-able in modal windows but since Adw.Dialogs are displayed in the window, they are (ref: #1055)

I don't know what the best approach would be right now :shrug:

toolstack:
My initial thought would be to disable the nav accels (home/pgup/pgdn) when opening the dialog and then re-enabling them when closing it.  Same for when the search field is focused/unfocused.

From the [gtk docs](https://docs.gtk.org/gtk3/method.Application.set_accels_for_action.html) for `set_accels_for_action`:
```
To remove all accelerators for an action, use an empty, zero-terminated array for accels.
```

Perhaps even disable all (or a broader subset at least) of the accels when opening a dialog as the desired behaviour is probably not to be able to use any of them when a dialog is already open (aka at the moment you can bring up the about box, then while it's open, bring up the preferences, through the shortcut keys)

toolstack:
As for the issues with pgup/pgdn not working when another area (like the sidebar) is focused, it could probably be handled with a check to see what is being focused when the accel fires... aka if its *not* something like the sidebar, then scroll the post list, but if it is the sidebar, scroll that instead.

toolstack:
> We can disable individual actions (`SimpleAction#set_enabled`) but either way this is not the best solution either. I didn't test it but I think _any_ text field would trigger it. Including the entries in the Search and Lists views.

Agreed, any widget that would normally use the nav keys would be impacted with the problem.

> The way it is right now, they work wherever the focus position is. Which is the desired behavior.

Mostly agree, though pgup/dn break inside of the edit field of the compose window as well, so while not as significant of problem as Home, it is still broken in the same way.

> Also just to point it out, Home works as expected when the focus is in the scroll view. What prompted you to make the initial PR was the desire to have it work outside of it, just like the page up/down ones (rightfully so, since the focus will almost never be in the scrollview if you are using a mouse (unless you click a post)):

Agreed, pgup/dn and home should have the same functionality overall with whatever solution is decided upon.

I think there are two overall options:
1. Use the global accels and then disable them were they shouldn't function (like search boxes, text input, dialog boxes, etc)
2. Don't use the global accels and instead use local keyboard capture on the various widgets where the functionality should be (aka on the containing windows and most common focus areas that don't use the nav keys) 

Neither is particularly good but both would get the functionality that I think everyone expects.

toolstack:
@GeopJr I've thrown together option 1, disabling the global accelerators as needed, over [here](https://github.com/toolstack/Tuba/tree/home-key) as a proof of concept, probably needs some thought/changes around if the accels for pgup/dn should still be active in text boxes like the new lists one, but demonstrates a reasonable solution to the issue I think.

PR #1271, opened 2024-12-14
fix(MediaViewer): Make button-hide-timeout match gtk and clapper

rmader:
Update the timeout to 3s which is used both by GtkVideo and ClapperGtkVideo.
Note that the later uses 5s for touch by default, but that could be adjusted via `set_fade_delay()`.

This makes the fullscreen button disappear together with other media controls and matches defaults of other players more closely, making the overall experience more consistant.


GeopJr:
LGTM, thanks!

PR #1283, opened 2024-12-16
feat(Windows): nsis

GeopJr:
Time to one up the portable windows builds by offering an exe installer.

A lot to do

- [x] CI succeeds
- [x] Installer branding
- [x] Test on Windows without msys
- [x] Test uninstaller
- [x] Test updating
- [x] Differenciate dev vs release builds
- [x] Release workflow?

The installer config is based on https://github.com/SpikedPaladin/ValaGtkTemplate/ - we already had the foundation for the everything else needed from the portable builds


LukaszH77:
Whoah, you are busy again o.O What happened to the last part of this TODO? xD

obraz

GeopJr:
i lied 

when-i-purposefully-spread-misinformation-over-the-internet

I would have taken a break but found some... outrageous bugs 💀 FediWrapped default style is set to transparent and re-scheduling does not keep track of the month (and if it did it would be wrong because the calendar goes from 0-11 instead of 1-12)

Look at the last attachment 😭 'scheduled for january 1 2025' -> reschedule -> scheduled for december 1 2025' https://floss.social/@Tuba/113657758389560981

On the bright side the windows installer is taking shape!

image

PR #1299, opened 2024-12-28
Avoid measuring labels width-for-height

bugaevc:
These were the remaining places in Tuba where labels were still being measured width-for-height (at least, on my timeline right now), and with these simple changes, we seem to get rid of them too. See the commit message.

There are likely some more small things here and there; for instance I saw a similarly start-aligned wrappable "By %s" button. That should be avoidable using the same valign=center trick.


GeopJr:
LGTM, thanks!

PR #1316, opened 2025-01-31
fix(ci): flatpak

GeopJr:
aarch64 => use clang
bump action container to 47

i dont think it will fix uploading


GeopJr:
clang was a success, now on to fixing the uploading part of the action

GeopJr:
Would love if we stopped breaking workflows by changing artifact names but whatever, not worth forking over.

Issue #1347, opened 2025-03-03
[Bug]: Crash when clicking hashtags when going back to a toot after having entered details view


nekohayo:
### Describe the bug

Mostly the same as issue #1340, but for hashtags, anywhere.

### Steps To Reproduce

1. Click on a toot to view it in detailed view
2. Go back to the previous view
3. Click on one of its hashtags

### Logs and/or Screenshots

```
Thread 1 "dev.geopjr.Tuba" received signal SIGSEGV, Segmentation fault.
0x00007ffff7d9c8bc in gee_abstract_collection_foreach () from /usr/lib/x86_64-linux-gnu/libgee-0.8.so.2
(gdb) bt
#0  0x00007ffff7d9c8bc in gee_abstract_collection_foreach () at /usr/lib/x86_64-linux-gnu/libgee-0.8.so.2
#1  0x00005555556a034e in tuba_widgets_rich_label_on_activate_link (self=0x555557b03260, url=<optimized out>) at ../src/Widgets/RichLabel.vala:121
#2  0x000055555568d54e in g_cclosure_user_marshal_BOOLEAN__STRING
    (n_param_values=2, invocation_hint=<optimized out>, closure=<optimized out>, return_value=0x7fffffffb800, param_values=<optimized out>, marshal_data=<optimized out>)
    at ../src/Widgets/LabelWithWidgets.vala:10
#3  g_cclosure_user_marshal_BOOLEAN__STRING
    (closure=<optimized out>, return_value=0x7fffffffb800, n_param_values=<optimized out>, param_values=<optimized out>, invocation_hint=<optimized out>, marshal_data=<optimized out>) at /run/build/tuba/_flatpak_build/LabelWithWidgets.c:2516
#4  0x00007ffff7d3691a in g_closure_invoke () at /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#5  0x00007ffff7d4b5c3 in signal_emit_unlocked_R.isra.0 () at /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#6  0x00007ffff7d4c8f2 in signal_emit_valist_unlocked () at /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#7  0x00007ffff7d52e01 in g_signal_emit_valist () at /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
[...]

Issue #1340, opened 2025-02-24
[Bug]: Crash when clicking the avatar of the post at the top of the timeline, after having entered details view and gone back to the view


nekohayo:
### Describe the bug

I have devised a novel way to do this:

![Image](https://github.com/user-attachments/assets/d0fb8b52-99d6-4051-b42a-12b9f240c45e)

### Steps To Reproduce

1. ~~Click on Notifications~~
2. Click on a reply mentioning you / the 1st toot at the top of the timeline
3. Go back to the ~~notifications~~ previous page (either using the mouse's dedicated Back button, or clicking the `<` button on the headerbar)
4. Click on the avatar of the person mentioning you
5. Boooooooommm

### Logs and/or Screenshots

With `flatpak run --devel --command=sh dev.geopjr.Tuba` then `gdb dev.geopjr.Tuba`, I was able to capture this:

```
Thread 1 "dev.geopjr.Tuba" received signal SIGSEGV, Segmentation fault.
0x0000555555588c4d in tuba_api_account_get_id (self=self@entry=0x555557e328b0) at ../src/API/Account.vala:5
5		public string id { get; set; }
(gdb) bt full
#0  0x0000555555588c4d in tuba_api_account_get_id (self=self@entry=0x555557e328b0) at ../src/API/Account.vala:5
        result = <optimized out>
        _tmp0_ = <optimized out>
#1  0x00005555555ab3c3 in tuba_api_relationship_construct_for_account (object_type=Python Exception <class 'gdb.error'>: No type named TypeNode.
, acc=0x555557e328b0) at ../src/API/Relationship.vala:36
        self = 0x0
        _tmp0_ = <optimized out>
        _tmp1_ = <optimized out>
        self = <optimized out>
        _tmp0_ = <optimized out>
        _tmp1_ = <optimized out>
        _g_boolean_var_11 = <optimized out>
#2  tuba_api_relationship_new_for_account (acc=acc@entry=0x555557e328b0) at ../src/API/Relationship.vala:35
#3  0x000055555565e5d4 in tuba_views_profile_profile_account_construct (object_type=Python Exception <class 'gdb.error'>: No type named TypeNode.
, t_acc=0x555557e328b0) at ../src/Views/Profile.vala:7
        self = 0x0
        _tmp0_ = <optimized out>
        _tmp1_ = <optimized out>
[...]

PR #1354, opened 2025-03-07
feat(Widgets): FadeBin

GeopJr:
One of these days I'll make libgeopjr or something because it's a shame for all these custom widgets to not be used in other apps.

Anyway, faded view. Not sure if we should go forward with this but it's configurable enough to be able to auto-reveal with a setting or similar.

https://github.com/user-attachments/assets/19a02fde-db3e-40a5-b35c-25d2c8ef8d32

It could also come with a configurable max-height but for now it's a constant.

### DIFF
--- a/src/Widgets/FadeBin.vala
+++ b/src/Widgets/FadeBin.vala
@@ -1,0 +1,150 @@
+public class Tuba.Widgets.FadeBin : Gtk.Widget {
+	const int MAX_HEIGHT = 300;
+	const float FADE_HEIGHT = 125f;
+
+	private Gtk.Widget? _child = null;
+	public Gtk.Widget? child {
+		get { return _child; }
+		set {
+			if (_child != null) _child.unparent ();
+			_child = value;
+			_child.set_parent (this);
+		}
+	}
+
+	~FadeBin () {
+		if (this.child != null) this.child.unparent ();
+	}
+
+	public override Gtk.SizeRequestMode get_request_mode () {
+		if (this.child != null) return this.child.get_request_mode ();
+		return Gtk.SizeRequestMode.HEIGHT_FOR_WIDTH;
+	}
+
+	private bool _reveal = false;
+	public bool reveal {
+		get { return _reveal; }
+		set {
+			if (_reveal != value) {
+				_reveal = value;
+				this.queue_draw ();
+				this.queue_resize ();
+			}
+		}
+	}
+
+	private bool _should_fade = false;
+	private bool should_fade {
+		get { return _should_fade; }
+		set {
+			if (_should_fade != value) {
+				_should_fade = value;
+				this.notify_property ("faded");
+			}
+		}
+	}
+
+	public bool faded {
+		get {
+			return this.should_fade && !reveal;
+		}
+	}
+
+	const Gsk.ColorStop[] GRADIENT = {
+		{ 0f, { 0, 0, 0, 1f } },
+		{ 1f, { 0, 0, 0, 0f } },
+	};
+
+	public override void size_allocate (int width, int height, int baseline) {
+		if (this.child == null) {
+			base.size_allocate (width, height, baseline);
+			this.should_fade = false;
+			return;
+		}
+
+		this.child.allocate (width, height, baseline, null);
+		if (this.reveal) {
+			this.should_fade = false;
+			return;
+		}
+
+		this.should_fade = this.child.get_height () >= MAX_HEIGHT;
+	}
+
+	public override void measure (Gtk.Orientation orientation, int for_size, out int minimum, out int natural, out int minimum_baseline, out int natural_baseline) {
+		this.child.measure (
+			orientation,
+			for_size,
+			out minimum,
+			out natural,
+			out minimum_baseline,
+			out natural_baseline
+		);
+		if (this.reveal) return;
+
+		minimum = int.min (minimum, MAX_HEIGHT);
+		minimum_baseline = -1;
+		natural_baseline = -1;
+		if (orientation == Gtk.Orientation.VERTICAL) {
+			natural = MAX_HEIGHT;
+		}
+	}
+
+	public override void snapshot (Gtk.Snapshot snapshot) {
+		if (this.child == null || !this.faded) {
+			this.child.snapshot (snapshot);
+			return;
+		}
+
+		var height = this.get_height ();
+		if (height <= 0) {
+			this.child.snapshot (snapshot);
+			return;
+		}
+
+		Gtk.Snapshot child_snapshot = new Gtk.Snapshot ();
+		this.snapshot_child (this.child, child_snapshot);
+		var node = child_snapshot.to_node ();
+		if (node == null) {
+			this.child.snapshot (snapshot);
+			return;
+		}
+
+		var bounds = node.get_bounds ();
+		bounds.origin.y = 0;
+		bounds.origin.x = Math.floorf (bounds.origin.x);
+		bounds.size.width = Math.ceilf (bounds.size.width) + 1;
+		bounds.size.height = height;
+
+		snapshot.push_mask (Gsk.MaskMode.INVERTED_ALPHA);
+
+		var new_fade = height - FADE_HEIGHT;
+		snapshot.append_linear_gradient (
+			Graphene.Rect () {
+				origin = Graphene.Point () {
+					x = bounds.origin.x,
+					y = new_fade
+				},
+				size = Graphene.Size () {
+					width = bounds.size.width,
+					height = FADE_HEIGHT
+				}
+			},
+			Graphene.Point () {
+				x = 0,
+				y = height
+			},
+			Graphene.Point () {
+				x = 0,
+				y = new_fade
+			},
+			GRADIENT
+		);
+
+		snapshot.pop ();
+		snapshot.push_clip (bounds);
+		snapshot.append_node (node);
+		snapshot.pop ();
+		snapshot.pop ();
+	}
+}
--- a/src/Widgets/FadeBin.vala
+++ b/src/Widgets/FadeBin.vala
@@ -80,21 +80,25 @@ public class Tuba.Widgets.FadeBin : Gtk.Widget {
 			return;
 		}
 
+		int child_for_size;
+		if (this.reveal || orientation == Gtk.Orientation.VERTICAL || for_size < MAX_HEIGHT) {
+		    child_for_size = for_size;
+		} else {
+		    child_for_size = -1;
+		}
+
 		this.child.measure (
 			orientation,
-			for_size,
+			child_for_size,
 			out minimum,
 			out natural,
 			out minimum_baseline,
 			out natural_baseline
 		);
-		if (this.reveal) return;
 
-		minimum_baseline = -1;
-		natural_baseline = -1;
-		if (orientation == Gtk.Orientation.VERTICAL) {
-			minimum = int.min (minimum, MAX_HEIGHT);
-			natural = int.min (natural, MAX_HEIGHT);
+		if (orientation == Gtk.Orientation.VERTICAL && !this.reveal) {
+		    minimum = int.min (minimum, MAX_HEIGHT);
+		    natural = int.min (natural, MAX_HEIGHT);
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


GeopJr:
@bugaevc quick question, would this have any impact on hitting the fast path?

FadeBin is HEIGHT_FOR_WIDTH unless there's a child, which then uses the child's mode. In statuses, there's always a child, the markupview.

The rest is mainly setting the min (<=) and nat (vertical) sizes to 300. Allocation passes to the child.

src/Widgets/FadeBin.vala
```diff
@@ -0,0 +1,150 @@
+public class Tuba.Widgets.FadeBin : Gtk.Widget {
+	const int MAX_HEIGHT = 300;
+	const float FADE_HEIGHT = 125f;
+
+	private Gtk.Widget? _child = null;
```
bugaevc, line 5: `unowned`

src/Widgets/FadeBin.vala
```diff
@@ -0,0 +1,150 @@
...
+	public Gtk.Widget? child {
+		get { return _child; }
+		set {
+			if (_child != null) _child.unparent ();
+			_child = value;
+			_child.set_parent (this);
```
bugaevc, line 11: null check?

src/Widgets/FadeBin.vala
```diff
@@ -0,0 +1,150 @@
...
+		}
+	}
+
+	~FadeBin () {
+		if (this.child != null) this.child.unparent ();
```
bugaevc, line 16: This should be done in [`dispose`](https://valadoc.org/gobject-2.0/GLib.Object.dispose.html), something like this:

```vala
public override void dispose () {
    this.child = null;
    base.dispose ();
}
```

src/Widgets/FadeBin.vala
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	public override Gtk.SizeRequestMode get_request_mode () {
+		if (this.child != null) return this.child.get_request_mode ();
+		return Gtk.SizeRequestMode.HEIGHT_FOR_WIDTH;
```
bugaevc, line 21: If you don't have a child, return `CONSTANT_SIZE`.

src/Widgets/FadeBin.vala
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	private bool _reveal = false;
+	public bool reveal {
+		get { return _reveal; }
+		set {
+			if (_reveal != value) {
+				_reveal = value;
+				this.queue_draw ();
+				this.queue_resize ();
```
bugaevc, line 31: `queue_resize ()` implies `queue_draw ()`, no point in calling both

src/Widgets/FadeBin.vala
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	const Gsk.ColorStop[] GRADIENT = {
+		{ 0f, { 0, 0, 0, 1f } },
+		{ 1f, { 0, 0, 0, 0f } },
+	};
+
+	public override void size_allocate (int width, int height, int baseline) {
+		if (this.child == null) {
+			base.size_allocate (width, height, baseline);
```
bugaevc, line 60: The default implementation does nothing, so normally nobody bothers to call it.

src/Widgets/FadeBin.vala
```diff
@@ -0,0 +1,150 @@
...
+			return;
+		}
+
+		this.child.allocate (width, height, baseline, null);
+		if (this.reveal) {
+			this.should_fade = false;
+			return;
+		}
+
+		this.should_fade = this.child.get_height () >= MAX_HEIGHT;
```
bugaevc, line 71: Don't use `get_height ()` like that, it's for the widget itself to call. Use `get_allocated_height ()` (which is deprecated I think?), or better, `height`, since that's what you've just allocated.

EDIT: `child_height` not `height`, if you follow the suggestion above.

src/Widgets/FadeBin.vala
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	public override void measure (Gtk.Orientation orientation, int for_size, out int minimum, out int natural, out int minimum_baseline, out int natural_baseline) {
+		this.child.measure (
```
bugaevc, line 75: null check?

src/Widgets/FadeBin.vala
```diff
@@ -0,0 +1,150 @@
...
+			out minimum_baseline,
+			out natural_baseline
+		);
+		if (this.reveal) return;
+
+		minimum = int.min (minimum, MAX_HEIGHT);
+		minimum_baseline = -1;
+		natural_baseline = -1;
+		if (orientation == Gtk.Orientation.VERTICAL) {
+			natural = MAX_HEIGHT;
```
bugaevc, line 89: `int.min (natural, MAX_HEIGHT)` rather

src/Widgets/FadeBin.vala
```diff
@@ -0,0 +1,150 @@
...
+			orientation,
+			for_size,
+			out minimum,
+			out natural,
+			out minimum_baseline,
+			out natural_baseline
+		);
+		if (this.reveal) return;
+
+		minimum = int.min (minimum, MAX_HEIGHT);
```
bugaevc, line 85: Only for `VERTICAL`, so move this into the `if` below.

src/Widgets/FadeBin.vala
```diff
@@ -0,0 +1,150 @@
...
+		minimum_baseline = -1;
+		natural_baseline = -1;
+		if (orientation == Gtk.Orientation.VERTICAL) {
+			natural = MAX_HEIGHT;
+		}
+	}
+
+	public override void snapshot (Gtk.Snapshot snapshot) {
+		if (this.child == null || !this.faded) {
+			this.child.snapshot (snapshot);
```
bugaevc, line 95: Again this looks wrong wrt null.

Also: never use `child.snapshot (snapshot)`; it's wrong of Vala to even expose it (file a bug, perhaps). Always use `snapshot_child (child, snapshot)`.

GeopJr:
Thanks for the reviews! If you conclude that this is not feasible in a performant way, I'm okay with dropping it.

src/Widgets/FadeBin.vala
GeopJr: Honestly, I'm scared of using unowned / weak ([`At the moment weak and unowned can be used interchangeably`](https://wiki.gnome.org/Projects/Vala/ReferenceHandling)) anymore. Every segfault Tuba has had is due to using them because, unless there's some secret way to do this that I don't know of, null checking is not enough and will happily pass them if it has been freed and then segfault.

Latest #1340 #1347 (also, both cases are very weird since the status should still hold a reference to them anyway!).

I guess it won't hurt testing it out and see if we hit it.

src/Widgets/FadeBin.vala
```diff
@@ -0,0 +1,150 @@
...
+			return;
+		}
+
+		Gtk.Snapshot child_snapshot = new Gtk.Snapshot ();
+		this.snapshot_child (this.child, child_snapshot);
+		var node = child_snapshot.to_node ();
+		if (node == null) {
+			this.child.snapshot (snapshot);
+			return;
+		}
```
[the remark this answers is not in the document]
GeopJr: The quoted snippet is for when there are no nodes. The snapshot_child before it, is for getting the node bounds. I can change it to free_to_node as we have no use for it afterwards, but we do need the node bounds for gradient.

bugaevc:
I do think it is both feasible and interesting, please keep going :slightly_smiling_face:

src/Widgets/FadeBin.vala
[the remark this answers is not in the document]
bugaevc: You shouldn't need the node bounds either, something like this should work:

```vala
var width = get_width ();
var height = get_height ();

snapshot.push_mask (Gsk.MaskMode.INVERTED_ALPHA);
snapshot.append_linear_gradient (
  { { 0, height - FADE_HEIGHT }, { width, FADE_HEIGHT } },
  { 0, height },
  { 0, height - FADE_HEIGHT },
  GRADIENT
);

snapshot.pop ();
snapshot_child (this.child, snapshot);
snapshot.pop ();
```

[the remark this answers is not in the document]
GeopJr: Nope :(

image

src/Widgets/FadeBin.vala
```diff
@@ -0,0 +1,150 @@
...
+
+		this.should_fade = this.child.get_height () >= MAX_HEIGHT;
+	}
+
+	public override void measure (Gtk.Orientation orientation, int for_size, out int minimum, out int natural, out int minimum_baseline, out int natural_baseline) {
+		this.child.measure (
+			orientation,
+			for_size,
```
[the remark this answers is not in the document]
GeopJr: Thanks :heart:

src/Widgets/FadeBin.vala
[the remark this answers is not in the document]
bugaevc: Ah well, you should clip to your own bounds then. `{ 0, 0, width, height }` for starters, but you're probably only going to want to clip to the lower edge to keep things like shadows working otherwise.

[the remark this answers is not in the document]
GeopJr: I mean, give me a second to play around with clipping to the same rectangle.

src/Widgets/FadeBin.vala
```diff
@@ -80,21 +80,25 @@ public class Tuba.Widgets.FadeBin : Gtk.Widget {
...
+			child_for_size,
 			out minimum,
 			out natural,
 			out minimum_baseline,
 			out natural_baseline
 		);
-		if (this.reveal) return;
 
-		minimum_baseline = -1;
-		natural_baseline = -1;
```
bugaevc: I meant, keep setting these to `-1` for `!reveal` branch. Not that it matters.

GeopJr:
> I'd expect the fade/size to animate when expanded, like Gtk.Revealer does. Do you not want that?

It might be too much for very long posts but I'll see about it. If we go for it, I'd probably guess:
- snapshot whole child no fade
- overflow hidden
- set min/nat to `int.max (minimum * animation.value, MAX_HEIGHT);`

bugaevc:
> set min/nat to `int.max (minimum * animation.value, MAX_HEIGHT);`

It's a lot more complex than that. Basically, you want to animate the height limit from `MAX_HEIGHT` up to "child's actual height". Again, I would have to think for (a much longer) while about the correct way here. It's probably best if you don't block this PR on the animation, we can add it later.

GeopJr:
> we can add it later

:+1: 

This PR is not done anyway, it was an initial implementation. I still need to add settings and polish the UI, strings etc.

bugaevc:
Here's a working sketch of the animation
https://gitlab.gnome.org/-/snippets/6864

GeopJr:
Thanks! Is there a specific reason on doing the animation manually instead of using AdwTimedAnimation?

bugaevc:
Ah, sure, feel free to use that. I was focusing on the implementation of measure & size allocate, not on everything else.

GeopJr:
I don't think I skipped anything, but the main difference is that FadeBin's API needs a way to expand without animation for cases like users having collapsing long posts off and pre-expanding them when opening a thread, so there's the `reveal` property and the `reveal_animated` (and `hide_animated`) functions. When the animation ends, the reveal property inverts.

https://github.com/user-attachments/assets/3637f854-7be7-4d52-99e5-3b46ffcadb76

bugaevc:
* Does it work — does it animate nicely? Can you record a video? :smiley: Ideally, I should build this branch, slow down the animation, and verify that nothing about the layout invariants breaks.
* Not sure if the `LOTS` thing causes GSK to render a huge offscreen texture, or if it already does optimizations based on bounds etc.

bugaevc:
I see, cool :smile:

GeopJr:
> Not sure if the LOTS thing causes GSK to render a huge offscreen texture, or if it already does optimizations based on bounds etc.

I wonder if extreme edge cases will have any impact (100k+ character long posts) :thinking: Will test it out later.

Thanks again for everything! :bow:

PR #1355, opened 2025-03-09
fix(MainWindow): bind sidebar enable-gestures to `collapsed`

DaPigGuy:
On touchscreen devices, it is possible to hide the sidebar when it is not collapsed by using a touchscreen swipe gesture. The sidebar becomes inaccessible as a result. The gestures should only be enabled when the sidebar is collapsed (an overlay), where it is expected to be shown and hidden with the sidebar button.


GeopJr:
LGTM, thanks!

PR #1380, opened 2025-04-17
Use GApplication to parse command line options

City-busz:
It already supports adding and parsing option entries, so it's not necessary to process them manually.

Also, mark the description of the option entry as translatable.


GeopJr:
Thanks for this PR!

I'll get to this and the others as soon-ish, feel free to leave them as is, I'll merge main etc. when I get to them!

PR #1381, opened 2025-04-17
Add D-Bus service file

City-busz:
And mark the application as D-Bus activatable. This allows application launchers to activate it via D-Bus.

When the application is run in service mode, it's not supposed to show window, so I moved `accounts.init ()` from `startup` to `activate` phase and ensure that it runs only once, because it calls `ensure_active_account ()`, which in turn calls `present_window ()` if no account configured.

Reference:
https://specifications.freedesktop.org/desktop-entry-spec/1.3/dbus.html

Depends on https://github.com/GeopJr/Tuba/pull/1380


GeopJr:
:thinking: Do we really need this?

I get the point of it, but having the app run as a service sounds a bit too much IMO. The current solution of `--hidden`, uses (or it least it should) the background portal and will display Tuba in the background apps section of the DE, allowing users to see that 1. it's running and 2. kill it. This is what it looks like in GNOME:

image

> When the application is run in service mode, it's not supposed to show window, so I moved accounts.init () from startup to activate phase and ensured that it runs only once, because it calls ensure_active_account (), which in turn calls present_window () if no account configured.

Which I also think it's important. It should also show errors when Tuba can't access the keyring or the token has expired / been revoked.

City-busz:
This is not the same as `--hidden`. The GApplication service is just used to launch the application, and it terminates the process if no action given after 10 seconds.

It has some benefits as described here: https://wiki.gnome.org/HowDoI(2f)DBusApplicationLaunching.html#Why.3F

When `DBusActivatable=true` is set in the desktop file, the application launcher of GNOME Shell and KDE Plasma launches the application via D-Bus instead of `fork()`/`exec()` the executable, so the D-Bus daemon will control the process by setting up its environment, adding it to the correct cgroup, just like how usual D-Bus services work.

The D-Bus interface is provided by GApplication, just the service file is missing.

Rafostar:
D-Bus activation allows app to have persistent notifications: [link](https://developer.gnome.org/documentation/tutorials/notifications.html). I would imagine something like this especially important for the apps like Tuba, no? :smile: There are also other benefits of being D-Bus activatable so its useful even if you do not plan of running as service.

GeopJr:
Okay that makes it a lot more appealing, thanks! I thought notifications already worked like that :/

City-busz:
> 1. What kind of actions can get activated through Dbus? If it's all the action entries, many of them are used in notification buttons and **require** an active account to work. Additionally, the account-specific gsettings also depend on an active account existing.

The actions can be listed by the following command:
```
$ gdbus call --session \
  --dest dev.geopjr.Tuba \
  --object-path /dev/geopjr/Tuba \
  --method org.gtk.Actions.List
```
The available actions can be limited by moving the `add_action_entries()` from _startup_ to the _activate_ phrase. Of course, many of these actions are broken if the main window is not constructed, so it's probably better to not expose them in the _startup_ phrase. And if an action is desirable as an entry point to the application, then it could call `this.activate ()` before do anything to ensure that the main window is initialized and constructed, and then these actions can be advertised in the `.desktop` file so application launchers could provide direct access to them.

> 2. Frankly, I _want_ it to show the new account dialog if there's no account, regardless if it's being run with --hidden or from dbus - the app hasn't been set up yet and shouldn't run without a window.

`accounts.init` is still called as soon as the application is _activated_. It just moved out from the _startup_ phrase, because the `startup` is called even in the service mode. The _startup_ phrase is really just meant to expose the D-Bus API to access the actions and initialize the most essential parts.

> 3. Is there any way to detect if it was launched with dbus and shouldn't present a window? I'd rather just prevent presenting a window than ensuring there's an active account late.

It can be checked by the following condition in the startup phrase: `(this.flags & ApplicationFlags.IS_SERVICE) != 0)`. But once the application is being activated, it's no longer in service mode. So basically we just need to ensure that no window is shown in startup phrase, and do any interactive tasks in the `activate` phrase.

> Lastly, should this part of the docs concern Tuba further? Does it break `--hidden` when a user has set `dev.geopjr.Tuba --hidden` to run on startup?
> 
> > Since 2.40, applications that are not explicitly flagged as services or launchers (ie: neither g_application_is_service or g_application_is_launcher are given as flags) will check (from the default handler for local_command_line) if "--gapplication-service" was given in the command line. If this flag is present then normal commandline processing is interrupted and the g_application_is_service flag is set. This provides a "compromise" solution whereby running an application directly from the commandline will invoke it in the normal way (which can be useful for debugging) while still allowing applications to be D-Bus activated in service mode. The D-Bus service file should invoke the executable with "--gapplication-service" as the sole commandline argument. This approach is suitable for use by most graphical applications but should not be used from applications like editors that need precise control over when processes invoked via the commandline will exit and what their exit status will be.

No, it does not break, because when `--hidden` is specified, the application is activated, so it won't enter into service mode.

City-busz:
Yes, it looks good.

PR #1408, opened 2025-05-12
feat(Iceshrimp): drive

GeopJr:
~~blocked atm: https://issues.iceshrimp.dev/issue/ISH-792/Iceshrimp-API-does-not-accept-mastoapi-token~~

https://github.com/user-attachments/assets/d2d2c949-0215-44df-a0fb-c710eb9d1ade

close: #1444


pancakesmeow:
The blocked issue won't be fixed by Iceshrimp.NET but a middle-ground solution was just merged. See https://iceshrimp.dev/iceshrimp/Iceshrimp.NET/pulls/195 for details.

GeopJr:
Sounds good to me, thanks for pushing this forward!

GeopJr:
@pancakesmeow quick question, is the IS token valid until revoked or is it per session?

I'm asking so I know if I should save it with the other account data in the keyring or just get a new one on every startup. (Specifically because on the MR you are requesting a 'WebSession')

PR #1418, opened 2025-05-15
allow raw ip, v6 including

d47081:
minor correction to allow IPv6

not sure if IP validation by the dot is really wanted here.
if that is host/name validation it could be improved with Glib API, for example [is_valid](https://docs.gtk.org/glib/type_func.Uri.is_valid.html) method instead or maybe something better exists for these needs. removed so as uneffective.

### DIFF
--- a/src/Dialogs/NewAccount.vala
+++ b/src/Dialogs/NewAccount.vala
@@ -137,7 +137,7 @@ public class Tuba.Dialogs.NewAccount: Adw.Window {
 		account.instance = str;
 		instance_entry.text = str;
 
-		if (str.char_count () <= 0 || !("." in account.instance))
+		if (str.char_count () <= 0)
 			throw new Oopsie.USER (_("Please enter a valid instance URL"));
 	}
 


GeopJr:
Please use the main branch! #1410 should already support ipv6 if you pass `TUBA_SKIP_STRICT_VALIDATION=1`. I'll delete the experiment branch to avoid confusion.

https://github.com/GeopJr/Tuba/blob/main/src/Dialogs/NewAccount.vala#L133-L176

It first checks if the env var is set to '1'
`bool skip_strict_validation = GLib.Environment.get_variable ("TUBA_SKIP_STRICT_VALIDATION") == "1";`

Then sets `https` as the URI scheme if it doesn't have one
`if (!final_string.contains ("://")) final_string = @"https://$final_string";`

Then does regular URI parsing (ipv6 should be valid), but if `skip_strict_validation` is false, it forces the scheme to be `https`, removes the port + userinfo and checks if dots exist in the host (<- this is what your PR would change)
```vala
if (!skip_strict_validation) {
  scheme = "https";
  port = -1;
  userinfo = null;

  if (!host.contains (".")) {
    throw new Error.literal (-1, 1, @"Host '$host' is missing a dot");
  }
}
```

Then it converts it back to a string for `final_string_no_scheme` but without the scheme. This is because we want to display it in the text entry field in a pretty format. However when the scheme is empty, `GLib.Uri#to_string` will return it like this `://<host><port>...`, so we need to remove the first 3 characters `.substring (3);`

Then we need to re-add the scheme ourselves, this will be used internally
`final_string = @"$scheme://$final_string_no_scheme";`

And finally set them:
```vala
account.instance = final_string; // <- includes scheme
instance_entry.text = final_string_no_scheme; // <- will be displayed in the entry field without scheme
```

Issue #1407, opened 2025-05-12
Can't update default post language (settings dialog)


d47081:
### Describe the bug

It getting reset back to English, when everything else saving well

### Steps To Reproduce

Preferences -> General tab -> Default post language

### Logs and/or Screenshots

![Image](https://github.com/user-attachments/assets/abfbe91e-5195-4ff6-8e14-9876b5d8b16c)

### Instance Backend

Mastodon

### Operating System

Linux

### Package

Flatpak

GeopJr:
Is your instance backend actually mastodon? That gets saved on your instance / not locally.

I can maybe add a fallback if it fails to store it locally.

FWIW, the network error should be printed in logs, run tuba with `G_MESSAGES_DEBUG=Tuba flatpak run dev.geopjr.Tuba` and reproduce it.

GeopJr:
Basically, most settings are local, which you can see here `flatpak run --command=gsettings dev.geopjr.Tuba list-recursively dev.geopjr.Tuba`, but 1-2 of them are saved in your server so they can be synced across clients. The default language and post visibility are 2 of them.

snac2 doesn't seem to implement the language one yet (https://codeberg.org/grunfink/snac2/src/commit/f321905d4bcdd4ae5b344f9b743ecdac35e3672f/mastoapi.c#L3511), let me open an issue. If you ran Tuba from the command line, it would print the error message probably.

GeopJr:
Tbh, I think it's a nice suggestion to save it locally if saving it to / retrieving it from the server fails

oooo-ps:
I made this patch: https://github.com/GeopJr/Tuba/commit/9220d7917ae8f36a158cf579aa15156220653cc8 ([branch](https://github.com/YGGverse/Tuba/tree/ukrainian-by-default))
even believe it could be implemented without hardcoded values.

PR #1420, opened 2025-05-16
Respect local language settings first

d47081:
Solution for #1407

### DIFF
--- a/src/Dialogs/Preferences.vala
+++ b/src/Dialogs/Preferences.vala
@@ -350,6 +350,9 @@ public class Tuba.Dialogs.Preferences : Adw.PreferencesDialog {
 	private void on_window_closed () {
 		if (lang_changed) {
 			var new_lang = ((Utils.Locales.Locale) default_language_combo_row.selected_item).locale;
+
+			settings.default_language = new_lang; // respect local settings first (#1407)
+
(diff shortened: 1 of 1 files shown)


GeopJr:
Not sure how to approach this... As mentioned in #1407 we save and load it from the server so it syncs across clients but I was under the assumption that snac2 would return null or an empty string since it doesn't support it but it seems that it always returns 'en':

```json5
{
    "id": "2d49d7e26c495b8ceea54d6011b5d44c",
    "username": "awoo",
    "acct": "awoo",
    "display_name": "awoo",
    "created_at": "2025-05-16T07:34:22Z",
    "last_status_at": "2025-05-16T07:34:22Z",
    "note": "",
    "url": "https://127.0.0.1/awoo",
    "locked": false,
    "bot": false,
    "emojis": [
    ],
    "avatar": "https://127.0.0.1/susie.png",
    "avatar_static": "https://127.0.0.1/susie.png",
    "header": "",
    "header_static": "",
    "source": {
        "privacy": "public",
        "language": "en", // <---
        "follow_requests_count": 0,
        "sensitive": false,
        "fields": [
        ],
        "note": "",
        "bot": false
    },
    "followers_count": 0,
    "following_count": 0,
    "statuses_count": 0
}
```

so there's no real way of having a fallback if every time we sync we get a valid language value :/

PR #1422, opened 2025-05-16
drop `TUBA_SKIP_STRICT_VALIDATION`, optimize Uri filter

d47081:
By https://github.com/GeopJr/Tuba/pull/1418

1. I think we can drop `TUBA_SKIP_STRICT_VALIDATION` as extra option may confuse new members.

2. Also, I've removed extra steps, looks working for me, could you please test it with Internet DNS, IPv4, etc?
I tested on raw IPv6/Yggdrasil + mastodon.social, and connected there successfully.

3. Not sure about `.substring (3)` removing, please keep in mind if it is really wanted

4. The scheme enforcement removed, because there is alternative protocols available, not HTTP only ([Gemini](https://github.com/dimkr/tootik/) for example)

5. is `instance_uri.get_userinfo ()` in use by ActivityProtocol or we can skip it also by the `null`?

### DIFF
--- a/src/Dialogs/NewAccount.vala
+++ b/src/Dialogs/NewAccount.vala
@@ -134,7 +134,6 @@ public class Tuba.Dialogs.NewAccount: Adw.Window {
 		string final_string = instance_entry.text;
 		if (!final_string.contains ("://")) final_string = @"https://$final_string";
 
-		string final_string_no_scheme = final_string;
 		try {
 			GLib.Uri instance_uri = GLib.Uri.parse (final_string, GLib.UriFlags.NONE);
 			final_string = GLib.Uri.build (


GeopJr:
I was typing this for #1418 but I'll post this here instead:

The point of `TUBA_SKIP_STRICT_VALIDATION` is to enforce a set of protections / standards by default while allowing cases that fall outside of them to still pass the newaccount dialog. The dot requirement is not for ipv4 but for domain names.

I get it that it's limiting and imposes arbitrary rules but fedi already has a bad rep for being too technical and I don't want to promote usage of insecure instances. Without e.g. forcing https, someone who has no knowledge on http or even how URIs are structured, could use `http://` out of muscle memory and cause all future connections to be http. Same thing 

Insecure fedi instances are a big :x: (excluding your use cases). All your messages, posts, images, search queries, sometimes even the links you click will be visible for all in your network and your ISP. I'd rather not be responsible for any of that.

I think, if anything we are doing too little. Mastodon for Android *searches* for instances as you type and unless it finds one, it won't allow you to add it.

> I think we can drop TUBA_SKIP_STRICT_VALIDATION as extra option may confuse new members.

It's meant for advanced users only

> Not sure about .substring (3) removing, please keep in mind if it is really wanted

Prior to your changes, `final_string_no_scheme` had an empty scheme so when calling `to_string` on it, it would return `://<host>`. With substring we would remove the `://` part.

`final_string_no_scheme` is used to replace the entry text and show the user what we really care about (the host), while keeping the scheme on `final_string` which we save and use for the API.

> The scheme enforcement removed, because there is alternative protocols available, not HTTP only

The gemini one doesn't even have mastoapi support plus I doubt libsoup supports it without some sort of proxy. Actually, now that I think about it, there's no point in supporting anything but HTTP as libsoup is an HTTP client, it won't magically support other protocols like gopher.

> is instance_uri.get_userinfo () in use by ActivityProtocol or we can skip it also by the null?

Without `TUBA_SKIP_STRICT_VALIDATION`, I don't want to support it at all but otherwise I don't think there's anything stopping anyone from using URI userinfo for their server :shrug: 

---

I understand that this is a bit disappointing but this makes on-boarding more technical and prone to errors than it needs to be for the sake of 0.01% of use-cases

zod076:
Hi, does this work for the Windows port ? Currently I can't connect using a raw Yggdrasil address.

GeopJr:
`TUBA_SKIP_STRICT_VALIDATION` was added recently and hasn't made it into a release yet. It should work on the nightlies https://github.com/GeopJr/Tuba?tab=readme-ov-file#nightly

PR #1423, opened 2025-05-17
feat: use meson features to control optional dependencies

mgorny:
Use the standard Meson `feature` options to control optional link-time dependencies, as documented in: https://mesonbuild.com/Build-options.html

More specifically, replace the current boolean options that are already present for some of the dependencies with `feature` type options, and add options for the remaining optional dependencies.  These options integrate with `dependency()` calls automatically, making the resulting code simpler and more reliable.  This also makes it possible to disable all the optional dependencies explicitly.

This fixes a bug that Meson would attempt to link to the dependencies (such as `clapper`) even if they were disabled, since disabling the options did not disable the dependency checks, and the dependencies were used in the `final_deps` unconditionally.

I've left `libwebp` as-is, that is without an option, since it is only used to emit a warning.  I have also removed a redundant `clapper` version check, since the current dependency is higher than that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,6 @@
 .PHONY: all install uninstall build test potfiles
 PREFIX ?= /usr
 
-clapper ?= 1
 msys_sys ?= mingw64
 # Remove the devel headerbar style:
 # make release=1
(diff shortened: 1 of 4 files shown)


mgorny:
The Windows builds are failing because they were building with `clapper` explicitly enabled, but the `clapper` dependency is not available. I can change the Makefile back to the old behavior by replacing `enabled` with `auto`, if that's desirable.

GeopJr:
LGTM, thanks!

- I made the clapper option deprecated from bool as some distros already use it (per the build options doc) (the other options either were never options or weren't options in any release yet)
- enabled all options for the flatpak as it's used in the CI and we need to test them
- dropped the clapper option from the makefile as auto is good enough

mgorny:
Thank you!

PR #1429, opened 2025-05-20
Some Vala things

bugaevc:
I'm surprised about the `g_type_ensure` thing. I remember [having to add one of those lines myself](https://github.com/GeopJr/Tuba/pull/405/files#r1271402561), so now I got the idea that I should teach Vala to just do it automatically, and turns out it already does, and everything seems to just work without them indeed.

### DIFF
--- a/src/Dialogs/MainWindow.vala
+++ b/src/Dialogs/MainWindow.vala
@@ -14,11 +14,6 @@ public class Tuba.Dialogs.MainWindow: Adw.ApplicationWindow, Saveable {
 
 	Views.Base? last_view = null;
 
-	static construct {
-		typeof (Views.MediaViewer).ensure ();
-		typeof (Views.Sidebar).ensure ();
-	}
-
 	construct {
 		construct_saveable (settings);
 
--- a/src/Dialogs/Preferences.vala
+++ b/src/Dialogs/Preferences.vala
@@ -172,10 +172,6 @@ public class Tuba.Dialogs.Preferences : Adw.PreferencesDialog {
 	private bool lang_changed { get; set; default=false; }
 	private bool privacy_changed { get; set; default=false; }
 
-	static construct {
-		typeof (ColorSchemeListModel).ensure ();
-	}
-
 	construct {
 		proxy_entry.text = settings.proxy;
 		post_visibility_combo_row.model = accounts.active.visibility_list;
--- a/src/Dialogs/ProfileEdit.vala
+++ b/src/Dialogs/ProfileEdit.vala
@@ -80,10 +80,6 @@ public class Tuba.Dialogs.ProfileEdit : Adw.Dialog {
 		name = _("All Supported Files")
 	};
 
-	static construct {
-		typeof (Widgets.CustomEmojiChooser).ensure ();
-	}
-
 	construct {
 		filter.add_mime_type ("image/jpeg");
 		filter.add_mime_type ("image/png");
--- a/src/Services/Network/Streamable.vala
+++ b/src/Services/Network/Streamable.vala
@@ -1,4 +1,4 @@
-public abstract interface Tuba.Streamable : Object {
+public interface Tuba.Streamable : Object {
 
 	public struct Event {
 		public string type;
--- a/src/Views/MediaViewer.vala
+++ b/src/Views/MediaViewer.vala
@@ -427,10 +427,6 @@ public class Tuba.Views.MediaViewer : Gtk.Widget, Gtk.Buildable, Adw.Swipeable {
 		}
 	}
 
-	static construct {
-		typeof (Widgets.ScaleRevealer).ensure ();
-	}
-
 	#if CLAPPER
 		string clapper_cache_dir;
 		Gee.HashMap<string, string> clapper_cached_urls;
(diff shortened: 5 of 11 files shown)


GeopJr:
> it already does

is this something done in a recent release? Because ~half year ago it was still a problem #1170

bugaevc:
No, it's been doing that since https://gitlab.gnome.org/GNOME/vala/-/commit/8f9979da7a75cb91a46e47e61bf35a41c8ff8083

Maybe there is a Heisenbug in Vala (that I could fix)? How do I reproduce that?

GeopJr:
> How do I reproduce that?

If I remember correctly, the common reproducer was opening the profile editor (open your profile => pencil headerbar button) without constructing a custom emoji picker prior (e.g. without opening the composer)

GeopJr:
For completeness, if Vala is supposed to generate `g_type_ensure`s in the same file, it just doesn't seem to do so:

left is with the type ensure commented out and that's pretty much the only change in the generated c code
image

bugaevc:
It does so in instance init (below in your screenshot), right before `gtk_widget_init_template`

GeopJr:
It doesn't seem to do so in `Vala 0.56.18`, the following is identical between the two files:

image

GeopJr:
[ProfileEdit.c.txt](https://github.com/user-attachments/files/20348883/ProfileEdit.c.txt)
[ProfileEdit_with_manual_ensure.c.txt](https://github.com/user-attachments/files/20348884/ProfileEdit_with_manual_ensure.c.txt)

GeopJr:
Not opposed to it as long as it doesn't create conflicts (Tuba.API.Account vs Tuba.Widgets.Account); we had enough of them in vala land (GLib.DateTime vs Tuba.DateTime)

bugaevc:
Okay, so I have a Vala branch that fixes this. Other than `Tuba.Dialogs.ProfileEdit` / `g_type_ensure (TUBA_WIDGETS_TYPE_CUSTOM_EMOJI_CHOOSER);`, the only type affected seems to be `Tuba.Dialogs.Preferences`, which now has:

```
+	g_type_ensure (TUBA_TYPE_COLOR_SCHEME_LIST_MODEL);
+	g_type_ensure (TUBA_TYPE_COLOR_SCHEME_LIST_ITEM);
+	g_type_ensure (TUBA_INSTANCE_ACCOUNT_TYPE_VISIBILITY);
+	g_type_ensure (TUBA_UTILS_LOCALES_TYPE_LOCALE);
+	g_type_ensure (TUBA_INSTANCE_ACCOUNT_TYPE_STATUS_CONTENT_TYPE);
```

Does that match your expectations, or do you know of more instances of missing `g_type_ensure`s?

P.S. https://floss.social/@bugaevc/114556715462967488, any idea?

GeopJr:
> Okay, so I have a Vala branch that fixes this. Other than Tuba.Dialogs.ProfileEdit / g_type_ensure (TUBA_WIDGETS_TYPE_CUSTOM_EMOJI_CHOOSER);, the only type affected seems to be Tuba.Dialogs.Preferences, which now has:

> Does that match your expectations, or do you know of more instances of missing g_type_ensures?

Awesome, thanks! Yes I think it's alright, the missing `g_type_ensure` on the Preferences dialog were never a problem so I didn't include them (all of those are used on startup in other parts).

> P.S. https://floss.social/@bugaevc/114556715462967488, any idea?

I think the main problem is that we do keep a reference to the main window under app, you can set it to null on shutdown and it should hit them:

```diff
diff --git a/src/Application.vala b/src/Application.vala
index d8ba5539..50add463 100644
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -391,6 +391,7 @@ namespace Tuba {
                        #endif
                        network.flush_cache ();
 
+                       app.main_window = null;
                        base.shutdown ();
                }
 
```

GeopJr:
Maybe I should write a more graceful shutdown process...

bugaevc:
I did this instead:

```diff
diff --git a/src/Application.vala b/src/Application.vala
index d8ba5539..8e2dc82e 100644
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -32,7 +32,7 @@ namespace Tuba {
 	public class Application : Adw.Application {
 
 		public GLib.ProxyResolver? proxy { get; set; default=null; }
-		public Dialogs.MainWindow? main_window { get; set; }
+		public unowned Dialogs.MainWindow? main_window { get; set; }
 		public Dialogs.NewAccount? add_account_window { get; set; }
 		public bool is_mobile { get; set; default=false; }
 		public bool is_online { get; private set; default=true; }
diff --git a/src/Dialogs/MainWindow.vala b/src/Dialogs/MainWindow.vala
index bf8e96c6..1ad4bbd6 100644
--- a/src/Dialogs/MainWindow.vala
+++ b/src/Dialogs/MainWindow.vala
@@ -261,4 +261,11 @@ public class Tuba.Dialogs.MainWindow: Adw.ApplicationWindow, Saveable {
 
 		last_view = view;
 	}
+
+	public override void dispose () {
+		if (app.main_window == this) {
+			app.main_window = null;
+		}
+		base.dispose ();
+	}
 }
```

Could've also used `WeakRef` I guess. But it works, thanks! (And indeed immediately runs into https://gitlab.gnome.org/GNOME/gtk/-/issues/5834 w/ my Vala patch to insert `gtk_widget_dispose_template` calls.)

PR #1434, opened 2025-05-21
Increased compose window width to accomodate the German translation

thonkdifferent:
In the current stable and nightly version of Tuba on the compose screen, when the system language is set to German (and possibly other languages with long words), the words on the center 3 buttons in the desktop (wide) configuration are reduce to ellipsis as seen below.

Bildschirmfoto vom 2025-05-21 18-47-39

This PR expands the width of the compose window to `630px` to fit these larger words, as seen below
Bildschirmfoto vom 2025-05-21 18-29-53

It also works well with English words
Bildschirmfoto vom 2025-05-21 18-19-00

It is worth noting that we cannot set the width property to simply be `-1` (auto-adjust) because GtkCenterBox as of the writing of this PR, it does not evenly split the padding between its 3 children. A proper fix would be a new layout subclass (EqualPaddingLayout/EqualPaddingCenterBox?) that would properly place the 3 widgets along the top side of the AdwDialog. However, this should suffice for now.

### DIFF
--- a/data/ui/dialogs/compose.ui
+++ b/data/ui/dialogs/compose.ui
@@ -3,7 +3,7 @@
 	<requires lib="gtk" version="4.0" />
 	<requires lib="libadwaita" version="1.0" />
 	<template class="TubaDialogsCompose" parent="AdwDialog">
-		<property name="content_width">500</property>
+		<property name="content_width">630</property>
 		<property name="content_height">400</property>
 		<property name="width_request">360</property>
 		<property name="height_request">200</property>


data/ui/dialogs/compose.ui
```diff
@@ -3,7 +3,7 @@
 	<requires lib="gtk" version="4.0" />
 	<requires lib="libadwaita" version="1.0" />
 	<template class="TubaDialogsCompose" parent="AdwDialog">
-		<property name="content_width">500</property>
+		<property name="content_width">630</property>
 		<property name="content_height">400</property>
```
GeopJr, line 7:
```suggestion
		<property name="content_height">504</property>
```

I guess we can keep the aspect ratio

GeopJr:
A proper fix IMO would be to either use shorter words in the German translation if possible or ellipsize the button, but tbh increasing the default size it's fine. I'm aiming to finish the new composer in this release (#553) so I don't know if this change will make it into a release :shrug: 

Thanks!

PR #1435, opened 2025-05-23
feat(clapper): prepare for 0.10

### LINKED ISSUE #1431: clapper_enhancer_check is deprecated
Hi, just friendly ping here.

https://rafostar.github.io/clapper/doc/clapper/func.enhancer_check.html

There's lots of undergoing changes to how Clapper enhancer plugin system works currently. So I am opening this as a heads up before Clapper 0.10 (still lots of work left to do).

Anyway, just `deprecated`, so it should still work as before and I implemented all things that this was doing underneath, so if Tuba still desires such "enhancer_check" it should be able to replicate its logic here relatively easy.

GeopJr:
fix: #1431


src/API/Status/PreviewCard.vala
```diff
@@ -121,6 +129,47 @@ public class Tuba.API.PreviewCard : Entity, Widgetizable {
...
+
+			var proxies = Clapper.get_global_enhancer_proxies ();
+			for (var i = 0; i < proxies.get_n_proxies (); i++) {
+				var proxy = proxies.get_proxy (i);
+
+				string? proxy_schemes = proxy.get_extra_data ("X-Schemes");
+				if (proxy_schemes == null || proxy_schemes == "") continue;
+
+				string[] schemes = proxy_schemes.split (";");
+				if (!(scheme in schemes)) continue;
```
Rafostar, line 148: This part can probably be simplified to something like:

```vala
if (!proxy.extra_data_lists_value ("X-Schemes", scheme)) continue;
```

`proxy_schemes.split (";");` - I not a Vala expert, but I imagine this underneath does `malloc` to create char pointers array and `malloc` on each string in it, while Clapper provided `proxy.extra_data_lists_value` iterates it without any split/malloc so it might be a tiny bit faster.

src/API/Status/PreviewCard.vala
```diff
@@ -121,6 +129,47 @@ public class Tuba.API.PreviewCard : Entity, Widgetizable {
...
+			if (!Clapper.WITH_ENHANCERS_LOADER) return false;
+			bool supported = false;
+
+			string? host = t_host;
+			if (host != null && host.has_prefix ("www.")) host = host.substring (4);
+
+			var proxies = Clapper.get_global_enhancer_proxies ();
+			for (var i = 0; i < proxies.get_n_proxies (); i++) {
+				var proxy = proxies.get_proxy (i);
+
```
Rafostar, line 143: Missing check whether proxy enhancer is of "Extractable" type, see: [proxy.target_has_interface](https://rafostar.github.io/clapper/doc/clapper/method.EnhancerProxy.target_has_interface.html)

src/API/Status/PreviewCard.vala
```diff
@@ -121,6 +129,47 @@ public class Tuba.API.PreviewCard : Entity, Widgetizable {
 		}
 	}
 
+	#if CLAPPER_0_10
+		private bool enhancer_check (string scheme, string? t_host = null) {
+			if (!Clapper.WITH_ENHANCERS_LOADER) return false;
```
Rafostar, line 134: Just for information, this is fine cause it might save you from calling `.substring (4)` below, so you can leave it as-is. But other than that it is perfectly safe for applications to call `Clapper.get_global_enhancer_proxies ()` - in this case returned `GListModel` will simply have zero elements and `for` loop would not execute.

src/API/Status/PreviewCard.vala
```diff
@@ -121,6 +129,47 @@ public class Tuba.API.PreviewCard : Entity, Widgetizable {
...
+				string? proxy_hosts = proxy.get_extra_data ("X-Hosts");
+				if (proxy_hosts == null || proxy_hosts == "") continue;
+
+				string[] hosts = proxy_hosts.split (";");
+				// both the host and the scheme
+				// match at this point
+				if (host in hosts) {
+					supported = true;
+					break;
+				}
```
Rafostar, line 166: Same as above, this can be probably simplified too:

```vala
if (proxy.extra_data_lists_value ("X-Hosts", host)) {
  supported = true;
  break;
}
```

I will update documentation later, with mention that its safe to call this even with "key" does not exist.

GeopJr:
> extra_data_lists_value

Ohhh, I didn't read the description, I just assumed it would only check the *whole* value and not treat them as list, so I dismissed it instantly 😔... Much better than splitting and all!

---

It doesn't look like it matters and it's too early to do optimizations since the API is not done, but I do wonder if it's worth keeping some sort of 'cache'.

E.g.

"https", "youtube.com": passed
"peertube", null: failed

Keep that in memory and next time those combinations of scheme and host are checked, instantly return whether it's supported from memory instead of checking all plugins again

src/API/Status/PreviewCard.vala
```diff
@@ -121,6 +129,47 @@ public class Tuba.API.PreviewCard : Entity, Widgetizable {
...
+		private bool enhancer_check (string scheme, string? t_host = null) {
+			if (!Clapper.WITH_ENHANCERS_LOADER) return false;
+			bool supported = false;
+
+			string? host = t_host;
+			if (host != null && host.has_prefix ("www.")) host = host.substring (4);
+
+			var proxies = Clapper.get_global_enhancer_proxies ();
+			for (var i = 0; i < proxies.get_n_proxies (); i++) {
+				var proxy = proxies.get_proxy (i);
```
Rafostar, line 142: I whitelisted "peek_proxy" for Vala in latest git, since it (Vala) understands the difference just fine.

Now its:

```vala
public Clapper.EnhancerProxy? get_proxy (uint index);
public unowned Clapper.EnhancerProxy? peek_proxy (uint index);
```

Not much, but it can save you unnecessary ref-unref cycle in this one case here (when iterating global list).

Rafostar:
> It doesn't look like it matters and it's too early to do optimizations since the API is not done, but I do wonder if it's worth keeping some sort of 'cache'.

I thought about this too, but unsure how much of a boost you would get here to be worth it. These strings are read during init, then stored in enhancer proxy object, so there is no string dup/free going on. Just saving the amount of time it takes to iterate (currently) up to 3 elements array.

GeopJr:
In regards of MPRIS moving to enhancers, is there anything I should do here?

Rafostar:
> In regards of MPRIS moving to enhancers, is there anything I should do here?

Built-in `ClapperFeature` objects are on their deprecation way in favour of being re-implemented as enhancer plugins. This work is currently still ongoing, so APIs for that are still subject to change. I am trying to keep compatibility, so even if your app does not update from using old (now deprecated) implementation it should still work.

GeopJr:
Sounds good to me and it makes sense to move it to enhancers! I'll add this to this PR eventually, after all this is a draft for 0.10 / I expect API changes :)

Rafostar:
> I expect API changes :)

Just dropping here links to migration guide(s) in case you ever decide to continue with this PR. No rush. Even with latest Clapper version, Tuba still compiles cleanly (only deprecations warnings). Cheers.

* [clapper-gtk-0-10](https://rafostar.github.io/clapper/doc/clapper-gtk/migrating-to-010.html)
* [clapper-0-10](https://rafostar.github.io/clapper/doc/clapper/migrating-to-010.html)

PR #1448, opened 2025-06-21
feat: add a setting to show sensitive media by default

accrawford0:
This adds a setting that can automatically show all sensitive media by default. This setting is off by default and functions independently from the Reveal Spoilers by Default setting. 

One motivation for adding this setting is that the spoilered-attachment CSS class is very slow (at least on lower end hardware).


GeopJr:
LGTM overall, thanks!

There's some edge cases that your change highlights that I want to fix before merging this, so I'll turn it into a draft. Specifically, has_spoiler is true when the post has a content warning for example; in that case the attachments should also be assumed to be sensitive (as mastodon does). The new setting has to take that into account which can be a bit messy. Then there's even more conditions when users toggle the sensitivity. And of course, 

https://github.com/GeopJr/Tuba/blob/9c0f7672fb56c2290ba760113e628e9330fc14f6/src/Widgets/Attachment/Box.vala#L148

this happens in update, which can be called when a post has been edited for example. It was definitely wrong before, it shouldn't reveal them when updating, but now it sparked the question, should it be revealed based on the setting when updating? What if you have this setting enabled, see some attachments you don't like, you hide them and then the post gets edited? It would re-show them based on the setting which sounds a bit unintended.

These widgets are from the Tootle era and are in need of a re-write eventually, so I wouldn't bother with restructuring the current state of things too much for these cases.

> One motivation for adding this setting is that the spoilered-attachment CSS class is very slow (at least on lower end hardware).

Is it actually the blur or something else? You can try this with the inspector to check:

```css
.spoilered-attachment {
	filter: none;
	opacity: 0;
}
```

Not that I doubt that the 30px blur is expensive but I wonder if it's a better way to deal with this or if something else is the bottleneck.

accrawford0:
I attempted to avoid updating the `_spoiler_revealed` boolean when a status is updated, but unfortunately it seems fairly impractical. When a status is updated, an entirely new instance of the Box class (containing the attachments) is made. In fact, through testing it appears the status object itself is deleted and reconstructed (and not rebound). So I'm not sure how to proceed with this.

GeopJr:
I can't remember off the top of my head the whole workflow (and as you noticed it's *very* nested for no good reason (which circles back to the whole thing needing to be rewritten :P)), but I'd probably say playing with the properties we already have at hand might be a good start.

Maybe something like:

- set the default value for `_spoiler_revealed` to `settings.show_sensitive_media`
- in the has_spoiler setter, replace `if (value) spoiler_revealed = false;` with `spoiler_revealed = spoiler_revealed` (yes, same value I'll explain below)
- and probably in update change it to `spoiler_revealed = spoiler_revealed;` (as above)

The initial state will always be the `settings.show_sensitive_media`. When `has_spoiler` gets set, which can happen after construction (? again, I don't remember the workflow from the top of my head), we set spoiler_revealed to the same value just to re-trigger its setter, now that we have the has_spoiler, which the spoiler_revealed setter depends on. And lastly on update do the exact same thing.

That way, the following scenario would play somewhat nicely:

User has show_sensitive_media set to true. A post shows up, it's attachments are not hidden => _spoiler_revealed == true.

The user doesn't like the content they saw so they manually hide the attachments => _spoiler_revealed == false.

The post gets updated and since _spoiler_revealed was set to false, the updated box will also keep the same state (not revealed).

---

I think that would do somewhat, though I'll have to test the edge cases.

TBH, I'd also want to get rid of the blur if it's that GPU intense. I don't think anyone will mind if the style changes to the one I mentioned before or similar. FWIW, the default renderer uses Vulkan (all GTK4 apps should), you can try the opengl one by running Tuba with `GSK_RENDERER=ngl` (will also work on all other GTK4 apps).

If you open the inspector (ctrl+shift+I), it should tell you which renderer is in use

image

If for some reason ngl is the default, try Vulkan (`GSK_RENDERER=vulkan`).

Unfortunately, I do not have access to these kinds of hardware so it's hard to optimize and debug performance issues on them from my side.

accrawford0:
> Maybe something like:

This looks like a great solution! It does work as intended, but unfortunately, the entire Box object is destroyed after a status edit (and it seems the status is too). So after a status edit, the entire status is effectively deleted and recreated, so even with your solution, after a message edit, the messages will be re-hidden or re-revealed. This also currently occurs in the main branch with text spoilers too unfortunately. However, your solution is certainly more elegant and would work if the architecture is redesigned.

> Unfortunately, I do not have access to these kinds of hardware so it's hard to optimize and debug performance issues on them from my side.

Unfortunately the A64 does not support Vulkan and its OpenGL ES version (2.0) is not supported by GTK4's ngl renderer anymore. So it must perform everything on its CPU.
However, I did test the vulkan and ngl renderers on a laptop with a Core i7-1260p and both resulted in about 15fps when scrolling over CSS blurred media.

> I don't think anyone will mind if the style changes to the one I mentioned before or similar.

I tested your BlurHash implementation in Tuba and it resulted in a similar performance to `opacity: 0` with CSS; it's very performant. With that said, BlurHash's blur is much stronger than the 30px blur. I wouldn't mind any of these solutions personally.

PR #1484, opened 2025-07-23
fix(FadeBin): measuring when line height is > 1

GeopJr:
FadeBin would measure incorrectly when line height was enabled.

This forces the HORIZONTAL measuring to be for_size -1, no idea if that breaks anything else :shrug:

### DIFF
--- a/src/Widgets/FadeBin.vala
+++ b/src/Widgets/FadeBin.vala
@@ -121,7 +121,7 @@ public class Tuba.Widgets.FadeBin : Gtk.Widget {
 		}
 
 		int child_for_size;
-		if (this.reveal || orientation == Gtk.Orientation.VERTICAL || for_size < MAX_HEIGHT_OVER || for_size == -1) {
+		if ((this.reveal || for_size < MAX_HEIGHT_OVER || for_size == -1) && orientation == Gtk.Orientation.VERTICAL) {
 			child_for_size = for_size;
 		} else if (this.animation.value == 0.0) {
 			child_for_size = -1;


bugaevc:
Huh?? When measuring vertically, we always want to propagate the `for_size` aka width, why not?

bugaevc:
And when measuring horizontally, we want to propagate that `for_size` in select cases (when revealed, when it would fit in the first place, when it's `-1`) instead of falling through to inverse lerping.

GeopJr:
It seems to have been caused by my #1426 changes, I think #1492 fixes it, I'll quote it here too:

> During the https://github.com/GeopJr/Tuba/pull/1426, I made it so it only fades when the height is MAX_HEIGHT + 100, to avoid cases were the height is like MAX_HEIGHT + 1 and the user revealing it ends up only revealing a single pixel of height. Then I made it so the lerp only runs when the sizes are > MAX_HEIGHT + 100 which ended up causing what https://github.com/GeopJr/Tuba/pull/1484 tried to fix.

GeopJr:
I still feel like I'm missing something on MAX_HEIGHT vs MAX_HEIGHT_OVER

bugaevc:
So the fun thing is I'm going to use FadeBin to explain layout things in a GUADEC talk the day after tomorrow... so we better make sure it actually works :sweat_smile:

Let me try to understand this, you render the child in full if it's shorter than `MAX_HEIGHT_OVER`, but if the child is `MAX_HEIGHT_OVER` or taller, then you collapse it to `MAX_HEIGHT` (and interpolate up from there during animation), is that right?

GeopJr:
> So the fun thing is I'm going to use FadeBin to explain layout things in a GUADEC talk the day after tomorrow... so we better make sure it actually works 😅

\*gulp\*

> Let me try to understand this, you render the child in full if it's shorter than MAX_HEIGHT_OVER, but if the child is MAX_HEIGHT_OVER or taller, then you collapse it to MAX_HEIGHT (and interpolate up from there during animation), is that right?

Yes exactly!

bugaevc:
Some more descriptive names like `HEIGHT_THRESHOLD` and `COLLAPSED_HEIGHT` would help.

Here's an attempt:

```vala
assert (COLLAPSED_HEIGHT < HEIGHT_THRESHOLD);

int child_for_size;
if (orientation == Gtk.Orientation.VERTICAL || this.reveal || for_size == -1 || for_size < COLLAPSED_HEIGHT) {
    child_for_size = for_size;
} else if (animation.value == 0.0) {
    child_for_size = -1;
} else {
    child_for_size = (int) Math.floor (inverse_lerp (COLLAPSED_HEIGHT, for_size, animation.value)));
    assert (child_for_size >= for_size);
}

child.measure (orientation, child_for_size, out minimum, out natural, out minimum_baseline, out natural_baseline);

if (orientation == Gtk.Orientation.VERTICAL && !this.reveal) {
    // ...baseline...
    if (minimum > HEIGHT_THRESHOLD) {
        minimum = (int) Math.ceil (lerp (COLLAPSED_HEIGHT, minimum, animation.value));
    }
    // same for natural
}
```

In case `COLLAPSED_HEIGHT <= for_size && for_size < HEIGHT_THRESHOLD`, then it could be that either:
* child size was less than `HEIGHT_THRESHOLD`, so the threshold never triggered (`child_for_size = for_size`); or
* child size was above the `HEIGHT_THRESHOLD`, and it got lerp'ed (`child_for_size = inverse_lerp (...)`).
Both cases could result in such a `for_size`, and we should pass the larger size as `child_for_size`. Since `inverse_lerp` should always return a larger value, that's what we pass.

Does that make sense / work?

bugaevc:
Hm, but actually it sounds this whole setup cannot work?

So you have a label which still fits below `HEIGHT_THRESHOLD` / `MAX_HEIGHT_OVER`, and so is shown in full. And then you make the FadeBin narrower, and the label becomes taller, and now it no longer fits into the height threshold, and so the FadeBin collapses it, and now it only needs `COLLAPSED_HEIGHT` / `MAX_HEIGHT` of height. So it shrunk along both orientations, which is not allowed.

bugaevc:
Ah well, I'm only going to be showing a screenshot & a simplified version of the logic, you don't have to revert anything. It's more about me being sure that I understand all the implications & subtleties myself, before I try to explain it to others, and it sounded like there was a bug that has slipped past me — but it's actually a generalization that you've been trying to add, which is not present in the version I'll be describing, so it's fine.

I have an idea about how this complex version could work, but I'd need to carefully write the various cases down, that's going to take some time, so likely after the talks :smile:

PR #1523, opened 2025-08-04
fix(meson): fix `gexiv2-0.16` lookup to respect `gexiv2` option

### LINKED ISSUE #1522: gexiv2-0.16 dependency does not respect `gexiv2` option
### Describe the bug

Per the following logic:

https://github.com/GeopJr/Tuba/blob/8db682e162f5699ec2d677d924877934b162a1ce/meson.build#L85-L88

If `gexiv2-0.16` is found, `gexiv2_dep` will be set and Tuba will be built with gexiv2 support enabled even if `-Dgexiv2=disabled` is passed.

### Steps To Reproduce

```
meson setup ${srcdir} -Dgexiv2=disabled
```

### Logs and/or Screenshots

_No response_

### Instance Backend

I'm not sure / Other

### Operating System

Gentoo Linux amd64

### Package

Compiled manually

_No response_

### Additional Context

_No response_

mgorny:
Pass both `gexiv2-0.16` and `gexiv2` packages to a single `dependency()` lookup, in order to simplify the logic and fix respecting `gexiv2` option.  This is supported since meson 0.60.0.  The minimum version can be left as `>=0.14`, since `gexiv2-0.16` will always be newer than `>=0.15`.  Thanks to Eli Schwartz for the suggestion.

Fixes #1522

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -82,10 +82,7 @@ clapper_gtk_dep = dependency('clapper-gtk-0.0', required: get_option('clapper'))
 gstreamer_dep = dependency('gstreamer-1.0', required: get_option('gstreamer'))
 webkit_dep = dependency('webkitgtk-6.0', required: get_option('in-app-browser'))
 
-gexiv2_dep = dependency('gexiv2-0.16', version: '>=0.15', required: false)
-if not gexiv2_dep.found ()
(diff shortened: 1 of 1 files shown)


GeopJr:
LGTM, thanks!

PR #1534, opened 2025-08-11
fix(translations): fix metainfo translations

a-wai:
Translations of the metainfo description are interpreted by libxml, therefore `&` must actually appear as `&amp;` in those. Fix the 3 languages where this is an issue.


a-wai:
Yes, I noticed the original text (from the metainfo.in) had `&amp;`, not sure how it's handled internally by gettext, but it likely converts it first.

Maybe replacing `&` in the original string with `and` would ensure this doesn't happen again? (I only noticed when building the Debian package, which runs tests by default)

PR #1539, opened 2025-08-15
fix(build): Tuba doesn't compile without libspelling feature enabled

Bnyro:
Hey, I noticed that the `update_spelling_lang ()` method is not found if the `LIBSPELLING` feature is disabled, so compilation without it would fail.


GeopJr:
Nice catch, thanks!

Issue #1540, opened 2025-08-15
Starting with v0.10.0 Tuba no longer attempts to unlock gnome-keyring if it is locked


nmeum:
### Describe the bug

I use Tuba with gnome-keyring, but without unlocking the keyring automatically on login through the PAM module. Starting with the v0.10.0 release, Tuba no longer prompts for the keyring to be unlocked on startup (if it is still locked) and instead displays the "New Account dialog":

```
$ G_MESSAGES_DEBUG=Tuba tuba
(tuba:21908): Tuba-DEBUG: 18:33:44.804: Application.vala:265: os: Alpine Linux (null)
(tuba:21908): Tuba-DEBUG: 18:33:44.804: Application.vala:265: prefix: /usr
(tuba:21908): Tuba-DEBUG: 18:33:44.804: Application.vala:265: flatpak: false
(tuba:21908): Tuba-DEBUG: 18:33:44.804: Application.vala:265: version: 0.10.1 (production)
(tuba:21908): Tuba-DEBUG: 18:33:44.804: Application.vala:265: gtk: 4.18.6 (4.18.6)
(tuba:21908): Tuba-DEBUG: 18:33:44.804: Application.vala:265: libadwaita: 1.7.6 (1.7.6)
(tuba:21908): Tuba-DEBUG: 18:33:44.804: Application.vala:265: libsoup: 3.6.5 (3.6.5)
(tuba:21908): Tuba-DEBUG: 18:33:44.804: Application.vala:265: libgtksourceview: 5.16.0 (5.16.0)
(tuba:21908): Tuba-DEBUG: 18:33:44.804: Application.vala:265: libspelling: true
(tuba:21908): Tuba-DEBUG: 18:33:44.804: Application.vala:265: Clapper: false
(tuba:21908): Tuba-DEBUG: 18:33:44.804: Application.vala:265: GStreamer: false
(tuba:21908): Tuba-DEBUG: 18:33:44.804: Application.vala:265: GExiv2: false
(tuba:21908): Tuba-DEBUG: 18:33:44.805: SecretAccountStore.vala:9: Using libsecret v0.21.7
(tuba:21908): Tuba-DEBUG: 18:33:44.807: SecretAccountStore.vala:94: Loaded 0 accounts
(tuba:21908): Tuba-DEBUG: 18:33:44.807: AccountStore.vala:83: Reset active account
[...]

PR #1541, opened 2025-08-19
fix(SAS): DONT_MATCH_NAME again

GeopJr:
fix: #1540

### DIFF
--- a/src/Services/Accounts/SecretAccountStore.vala
+++ b/src/Services/Accounts/SecretAccountStore.vala
@@ -11,9 +11,10 @@ public class Tuba.SecretAccountStore : AccountStore {
 		schema_attributes = new GLib.HashTable<string,Secret.SchemaAttributeType> (str_hash, str_equal);
 		schema_attributes["login"] = Secret.SchemaAttributeType.STRING;
 		schema_attributes["version"] = Secret.SchemaAttributeType.STRING;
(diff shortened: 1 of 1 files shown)


nmeum:
Can confirm that this fixes #1540.

src/Services/Accounts/SecretAccountStore.vala
```diff
@@ -11,9 +11,10 @@ public class Tuba.SecretAccountStore : AccountStore {
 		schema_attributes = new GLib.HashTable<string,Secret.SchemaAttributeType> (str_hash, str_equal);
 		schema_attributes["login"] = Secret.SchemaAttributeType.STRING;
 		schema_attributes["version"] = Secret.SchemaAttributeType.STRING;
+		//  schema_attributes["appid"] = Secret.SchemaAttributeType.STRING;
 		schema = new Secret.Schema.newv (
 			Build.DOMAIN,
-			Secret.SchemaFlags.NONE,
+			Secret.SchemaFlags.DONT_MATCH_NAME,
```
nmeum, line 17: Maybe it is worthwhile to add a comment with a [reference to the gnome-keyring upstream issue](https://gitlab.gnome.org/GNOME/gnome-keyring/-/issues/89) here so that the reader knows why the `DONT_MATCH_NAME` is necessary?

Issue #1551, opened 2025-09-17
Focus text field on new post button clicked


Sharkitty:
### Describe the request

When clicking the new post button, the text field of the newly opened subwindow isn't focused, so the user has to do a second click on that subwindow before they can type. It would be more userfriendly if this text field was automatically focused, so users can just press the button and start typing. Also, the reply button already has this behavior, so this would make the experience more consistent as well.

### Checks

- [x] This follows the [GNOME HIG](https://developer.gnome.org/hig/).

GeopJr:
Thanks, that's helpful! (It's not really a duplicate, because I thought I fixed it :'(). I'm starting to get a better idea of where the problem is, unfortunately I've been very busy this month and it might take me a bit to get back issue fixing.

Sharkitty:
Sorry, I was busy today, do you still need testing? I could find a moment for that tomorrow or Tuesday I think.

PR #1565, opened 2025-10-05
fix(composer): focus widget

GeopJr:
fix: #1551


GeopJr:
I'm positive that this is now fixed for good. My guess is - since I can't reproduce it - that it has to do with mapping and ticks adwdialog does. The editor is probably not mapped by the time it deals with the focus_widget on some devices. This PR now overrides its grab_focus to force our workflow (editor grab focus, scroller adjustment).

Issue #1581, opened 2025-10-18
attaching media file per drag and drop results in crash


FiSHYsolutions:
### Describe the bug

When I want to attach an image to a toot and drag and drop it into the post, Tuba (v0.10.3) crashes. If I attach the image to the post using the “Add media” button, it works without any problems.

### Steps To Reproduce

1. start a new post
2. attach a media file with drag and drop

### Logs and/or Screenshots

```
(dev.geopjr.Tuba:2836708): Gtk-CRITICAL **: 11:51:38.227: gtk_grid_query_child: assertion '_gtk_widget_get_parent (child) == (GtkWidget *) grid' failed

(dev.geopjr.Tuba:2836708): Gtk-CRITICAL **: 11:51:38.227: gtk_grid_remove: assertion 'gtk_widget_get_parent (child) == GTK_WIDGET (grid)' failed
**
ERROR:arraylist.c:735:gee_array_list_real_get: assertion failed: (index >= 0)
Bail out! ERROR:arraylist.c:735:gee_array_list_real_get: assertion failed: (index >= 0)
Abgebrochen (Speicherabzug geschrieben)
```

### Instance Backend

Mastodon

### Operating System

Ubuntu 22.04.5 LTS

### Package

Snap

GeopJr:
Can't reproduce here on snap, my guess is that it might have to do with the fact that the host is much older and/or snap. In the meantime, while I'm downloading 22.04 LTS, let me know if there are any relevant logs above the ones you sent!

GeopJr:
Confirmed snap and flatpak can access DnD files on 2404, either update to 2404 or report that apps cant access DnD files in sandboxes environments to ubuntu 🤷

PR #1582, opened 2025-10-18
fix(AttachmentsBin): race condition with adding attachments to the grid

GeopJr:
fix: #1581 

I assumed that by the time upload (async) starts, the attachment would have made it into the grid but the linked issue proved otherwise. When Tuba doesn't have access to the attachment\* it would reach an upload_error instantly and request for its removal from the grid, but it wasn't in it to begin with. It now adds it to the grid first.

\* this doesn't happen traditionally. If the user opens a file or DnD's it, it normally can access it or gets access to it even in sandboxed environments, but the linked issue was about the older Ubuntu LTS which has some issue with the DnD portal apparently and the app doesn't get access to the file then.


GeopJr:
Maybe check if it's parented instead on switch_place? I'll merge this for now but to be tested

PR #1583, opened 2025-10-19
Add Nynorsk translation

turtlegarden:
This pull request adds a Nynorsk translation to Tuba.[^1]

[^1]: I would have used Weblate, but Weblate was fussy and didn't want to create a Nynorsk translation file.


GeopJr:
Tried too, server error :(

LGTM, thanks!

Issue #1600, opened 2025-11-13
Windows app fails to open


dm30:
### Describe the bug

Hi. Amazing to see you took the time to build a Windows version. Unfotunately several libs are said to be missing when I try to open the just installed app.
Thank you so much.

### Steps To Reproduce

1. Install the app.
2. Click the icon to open it.
3. Several errors are prompted.

![Image](https://github.com/user-attachments/assets/5b02ceb8-860e-4e10-8ca8-191d4e9afe0d)

### Logs and/or Screenshots

```
<paste your logs here>
```

### Instance Backend

Mastodon

### Operating System

Windows 11 Pro

### Package

I'm not sure

freemin7:
Can reproduce on Windows 11 (non pro). The library "libssl-3-x64.dll" is missing for me too.
This affects both the installed and the portable version.  dev.geojpr.Tuba.exe does not produce any logs or command line output as fas as i can see.

SilkyPants:
Thanks for fixing these - though I tried downloading the artifacts from the latest Action, and get an error about missing `libicuuc78.dll`. 

Was something else updated and missed?

shreyas-malhotra:
Could you try installing a previous artifact (from before this PR) to confirm whether this is a regression introduced here, or an existing packaging issue?

dm30:
Unfortunately its still broken

shreyas-malhotra:
> Unfortunately its still broken

Where are you downloading the application from?

dm30:
I downloaded github's

shreyas-malhotra:
> I downloaded github's

Can you be a bit more specific, the latest release for the repository is from Oct 10, 2025 IST, the fix was pushed on Feb 22, 2026 IST, so if you're downloading the installer from releases page, it will be the same installer as the one you used earlier.

You'll have to run the installer binary given in the artifacts from the latest CI run to get the updated installer.

Please download the Tuba Setup.exe file from the artifacts section of the web page here, and then let me know if the issue still persists:

https://github.com/GeopJr/Tuba/actions/runs/22268319993

dm30:
Thank you. Sorry I didn't know. Now it works, although the window does not interact with the OS windows management, although I can access Mastodon

SilkyPants:
I'm just checking things on my end as well (Windows isn't my daily driver unfortunately) and I think I had downloaded the #1617 run artifacts instead.

https://github.com/GeopJr/Tuba/actions/runs/22268906736

[...]

PR #1634, opened 2026-02-22
Fix for Issue #1600 - Fixed Windows Installation - Windows: bundle MinGW runtime DLLs recursively (fixes missing OpenSSL dlls)

shreyas-malhotra:
The Windows portable build and NSIS installer were missing runtime DLLs provided by MinGW, specifically OpenSSL 3 (libcrypto-3-x64.dll, libssl-3-x64.dll). This caused the installer build to fail at runtime on clean Windows systems.

This change updates the Windows dependency bundling step to:
- Use ntldd -R (when available) to recursively resolve PE dependencies
- Copy all MinGW runtime DLLs into the portable directory
- Fall back to ldd if ntldd is not available (e.g. clangarm64)

No workflow changes were made. The fix is limited to the __windows_copy_deps target in the Makefile.

Results:
- OpenSSL runtime DLLs are now bundled correctly.
- The Windows installer launches on clean systems without requiring MSYS2 or a system OpenSSL installation.
- Reduces risk of future missing DLL issues due to indirect dependencies.

Fixes: #1600

Demonstration of Tuba running on Windows 11:

image


GeopJr:
LGTM overall, thanks!

For the sake of confirming that the CI also did it right, could you give the CI artifacts a try?

> The Windows installer launches on clean systems without requiring MSYS2 or a system OpenSSL installation.

(FWIW, as I said in that issue too, my windows environment is windows 10 iot and as it is now, the windows builds of Tuba worked as expected. I made sure multiple times by installing it without msys2 on the system at all and in VMs etc. I never tested on Windows 11 so I have no idea why it requires OpenSSL explicitly.)

shreyas-malhotra:
For the screenshot that I submitted with the PR, I installed Tuba using the artifact "Tuba Setup.exe" from the final CI run.

Tested the portable x86 binary as well just now, and it works as expected.

Couldn't check the portable aarch64 files though since I don't have an arm device running Windows.

GeopJr:
Sorry, one last question

> Add conditional check for ntldd command in Makefile, since ntldd was missing in workflow

Is it missing in the CI or not? If it is, we can just install it

shreyas-malhotra:
> Sorry, one last question
> 
> > Add conditional check for ntldd command in Makefile, since ntldd was missing in workflow
> 
> Is it missing in the CI or not? If it is, we can just install it

It was missing in the CI, so I wrote a simple check for it, if the check fails it runs an alternative command using ldd instead.

If we do install ntldd, best practice would then be to run CI again and verify that the build works with the ntldd command as well, or to simplify things we can just omit the ntldd command and the condition that checks for it, and just keep the revised ldd command in place, which is verified to work since ntldd wasn't present in the CI runs.

GeopJr:
hmm, I'm a bit confused here.

Prior to your changes, we already ran the exact same ldd command, which apparently missed OpenSSL for Windows 11. So if `ntldd` is missing in the CI, how come the CI artifact works on Windows 11 if technically nothing changed (since it ran the previous `ldd` command)?

GeopJr:
Don't worry about it, I'll just add it to the CI just to be safe

PR #1481, opened 2025-07-21
feat(Network): parse json async

GeopJr:
fix: #1480

PR #1643, opened 2026-03-24
experiment: requests v2

### LINKED ISSUE #1480: Hang in Tuba.Network.get_parser_from_inputstream ()
### Describe the bug

I'm sitting in a coffee shop, with a spotty network connection. Tuba seems to hang, and GNOME Shell offers me to kill it as unresponsive.

<details>
<summary>Here's a backtrace:</summary>
<pre><code>#0  0x00007ff5702406c2 in __syscall_cancel_arch () at /lib64/libc.so.6
#1  0x00007ff5702349da in __internal_syscall_cancel () at /lib64/libc.so.6
#2  0x00007ff570234a24 in __syscall_cancel () at /lib64/libc.so.6
#3  0x00007ff5702adc4e in poll () at /lib64/libc.so.6
#4  0x00007ff57204f46b in poll
    (__fds=<optimized out>, __nfds=<optimized out>, __timeout=<optimized out>)
    at /usr/include/bits/poll2.h:44
#5  0x00007ff571cdd5af in g_socket_condition_timed_wait
    (socket=<optimized out>, condition=<optimized out>, timeout_us=<optimized out>, cancellable=0x0, error=0x55ab9899c2d8) at ../gio/gsocket.c:4738
#6  0x00007ff571cdd92a in g_socket_receive_with_timeout
    (socket=0x55ab989063a0, buffer=0x55ab98a86b13, size=541, timeout_us=-1, cancellable=0x0, error=0x55ab9899c2d8) at ../gio/gsocket.c:3382
#7  0x00007ff571cbb296 in g_input_stream_read
    (stream=0x55ab988e7960, buffer=0x55ab98a86b13, count=541, cancellable=0x0, error=0x55ab9899c2d8) at ../gio/ginputstream.c:200
--Type <RET> for more, q to quit, c to continue without paging--
#8  0x00007ff53439d281 in g_tls_connection_gnutls_pull_func ()
    at /usr/lib64/gio/modules/libgiognutls.so
#9  0x00007ff56fc1304a in _gnutls_io_read_buffered () at /lib64/libgnutls.so.30
#10 0x00007ff56fc07cfc in _gnutls_recv_in_buffers () at /lib64/libgnutls.so.30
#11 0x00007ff56fc0a320 in _gnutls_recv_int () at /lib64/libgnutls.so.30
#12 0x00007ff5343a26df in g_tls_connection_gnutls_read ()
    at /usr/lib64/gio/modules/libgiognutls.so
#13 0x00007ff5343a502b in g_tls_connection_base_read ()
    at /usr/lib64/gio/modules/libgiognutls.so
#14 0x00007ff5343a55ec in g_tls_input_stream_read ()
    at /usr/lib64/gio/modules/libgiognutls.so
#15 0x00007ff571cbb296 in g_input_stream_read
    (stream=0x55ab97f38d70, buffer=0x55ab98936950, count=128, cancellable=0x0, error=0x7ffce171d3a0) at ../gio/ginputstream.c:200
#16 0x00007ff571e7dab4 in soup_filter_input_stream_read_fn ()
    at /lib64/libsoup-3.0.so.0
#17 0x00007ff571cbb296 in g_input_stream_read
    (stream=0x55ab9768c560, buffer=0x55ab98936950, count=128, cancellable=0x0, error=0x7ffce171d3a0) at ../gio/ginputstream.c:200
#18 0x00007ff571e80ba1 in soup_filter_input_stream_read_until ()
    at /lib64/libsoup-3.0.so.0
#19 0x00007ff571e5cbc0 in read_internal () at /lib64/libsoup-3.0.so.0
#20 0x00007ff571cbb296 in g_input_stream_read
--Type <RET> for more, q to quit, c to continue without paging--
    (stream=0x55ab97fe0d90, buffer=0x55ab98a91750, count=8192, cancellable=0x0, error=0x7ffce171d810) at ../gio/ginputstream.c:200
#21 0x00007ff571c8afcb in fill_input_buffer
    (stream=stream@entry=0x55ab98a8c260, at_least_size=at_least_size@entry=8192, blocking=blocking@entry=1, cancellable=cancellable@entry=0x0, error=error@entry=0x7ffce171d810) at ../gio/gconverterinputstream.c:363
#22 0x00007ff571c8b4a4 in read_internal
    (stream=0x55ab98a8c260, buffer=0x55ab98ad3b39, count=8192, blocking=1, cancellable=0x0, error=0x7ffce171d810) at ../gio/gconverterinputstream.c:426
#23 0x00007ff571cbb296 in g_input_stream_read
    (stream=0x55ab98a8c260, buffer=0x55ab98ad3b39, count=8192, cancellable=0x0, error=0x7ffce171d810) at ../gio/ginputstream.c:200
#24 0x00007ff571e7dab4 in soup_filter_input_stream_read_fn ()
    at /lib64/libsoup-3.0.so.0
#25 0x00007ff571e74257 in soup_client_input_stream_read_fn ()
    at /lib64/libsoup-3.0.so.0
#26 0x00007ff571cbb296 in g_input_stream_read
    (stream=0x55ab95c24bc0, buffer=0x55ab98ad3b39, count=8192, cancellable=0x0, error=0x7ffce171d810) at ../gio/ginputstream.c:200
#27 0x00007ff571c59aad in json_parser_load_from_stream ()
    at /lib64/libjson-glib-1.0.so.0
#28 0x000055ab790e93ec in tuba_network_get_parser_from_inputstream ()
#29 0x000055ab79125cfc in ___lambda139__tuba_network_success_callback ()
--Type <RET> for more, q to quit, c to continue without paging--
#30 0x000055ab790e9aa9 in ___lambda11__gasync_ready_callback ()
#31 0x00007ff571ce9e7c in g_task_return_now (task=0x55ab98798780)
    at ../gio/gtask.c:1363
#32 0x00007ff571cef590 in g_task_return
    (type=<optimized out>, task=0x55ab98798780) at ../gio/gtask.c:1432
#33 g_task_return (task=0x55ab98798780, type=<optimized out>)
    at ../gio/gtask.c:1389
#34 0x00007ff571e92e72 in async_send_request_return_result ()
    at /lib64/libsoup-3.0.so.0
#35 0x00007ff571e95ced in run_until_read_done.lto_priv ()
    at /lib64/libsoup-3.0.so.0
#36 0x00007ff571ce9e7c in g_task_return_now (task=0x55ab970c53a0)
    at ../gio/gtask.c:1363
#37 0x00007ff571cef590 in g_task_return
    (type=<optimized out>, task=0x55ab970c53a0) at ../gio/gtask.c:1432
#38 g_task_return (task=0x55ab970c53a0, type=<optimized out>)
    at ../gio/gtask.c:1389
#39 0x00007ff571e651c5 in io_run_until_read_async () at /lib64/libsoup-3.0.so.0
#40 0x00007ff571e65208 in io_run_until_read_ready () at /lib64/libsoup-3.0.so.0
#41 0x00007ff57203b880 in g_main_dispatch (context=0x55ab94f7f780)
    at ../glib/gmain.c:3398
#42 g_main_context_dispatch_unlocked (context=0x55ab94f7f780)
    at ../glib/gmain.c:4249
--Type <RET> for more, q to quit, c to continue without paging--
#43 0x00007ff5720447c8 in g_main_context_iterate_unlocked
    (context=context@entry=0x55ab94f7f780, block=block@entry=1, dispatch=dispatch@entry=1, self=<optimized out>) at ../glib/gmain.c:4314
#44 0x00007ff572044973 in g_main_context_iteration
    (context=context@entry=0x55ab94f7f780, may_block=may_block@entry=1)
    at ../glib/gmain.c:4379
#45 0x00007ff571d2202d in g_application_run
    (application=0x55ab951c5a50, argc=<optimized out>, argv=<optimized out>)
    at ../gio/gapplication.c:2715
#46 0x000055ab7904797e in main ()</code></pre>
</details>

As you can see, it's blocking on a synchronous network read on the main thread 😱. Specifically, `Tuba.Network.get_parser_from_inputstream ()` calls `Json.Parser.load_from_stream ()`, which reads the stream in a blocking manner.

### Steps To Reproduce

* Have a spotty network connection.
* Try
    ```
    (gdb) start
    (gdb) b soup_client_input_stream_read_fn thread 1
    ```
* Consider using something like https://gitlab.gnome.org/chergert/iobt to catch issues like this

### Logs and/or Screenshots

_No response_

### Instance Backend

Mastodon

### Operating System

Fedora Linux 42

### Package

OS repositories

_No response_

### Additional Context

_No response_

GeopJr:
sup: #1481
fix: #1480


src/Services/Network/Network.vala
```diff
@@ -106,6 +105,42 @@ public class Tuba.Network : GLib.Object {
...
+	) throws GLib.Error, Oopsie {
+		requests_processing++;
+
+		in_stream = yield session.send_async (msg, 0, cancellable);
+		var status = msg.status_code;
+		response_headers = msg.response_headers;
+
+		if (status >= 200 && status < 300) {
+			return true;
+		} else if (status == GLib.IOError.CANCELLED) {
```
bugaevc: Wait, why are you expecting HTTP status to be `19`? (Perhaps you're misreading the [migration guide](https://libsoup.gnome.org/libsoup-3.0/migrating-from-libsoup-2.html#status-codes-no-longer-used-for-internal-errors)?)

I'd expect the `session.send_async` call to throw a `GLib.IOError.CANCELLED` if you cancel the `cancellable`.

src/Services/Network/Network.vala
```diff
@@ -106,6 +105,42 @@ public class Tuba.Network : GLib.Object {
...
+			debug ("Message is cancelled.");
+		} else {
+			string error_msg = msg.reason_phrase;
```
bugaevc:
```suggestion
			unowned string error_msg = msg.reason_phrase;
```

src/Services/Network/Network.vala
```diff
@@ -106,6 +105,42 @@ public class Tuba.Network : GLib.Object {
...
+
+			try {
+				var parser = Network.get_parser_from_inputstream (in_stream);
+				var root = network.parse (parser);
```
bugaevc: Still blocking here?

If this is just intended and you're going to be moving to the async version is the other PR, please leave a warning comment.

src/Services/Network/Network.vala
```diff
@@ -106,6 +105,42 @@ public class Tuba.Network : GLib.Object {
...
+					? root.get_string_member_with_default ("message", msg.reason_phrase)
+					: root.get_string_member_with_default ("error", msg.reason_phrase);
+				}
+			} catch {}
+
+			critical (@"Request \"$(msg.uri.to_string ())\" failed: $status $(msg.reason_phrase) $error_msg");
+			throw new Oopsie.INSTANCE (error_msg);
+		}
+
+		return false;
```
bugaevc: Presumably we only reach here in the (wrong) `status == GLib.IOError.CANCELLED` branch?

src/Services/Network/RequestV2.vala
```diff
@@ -0,0 +1,158 @@
...
+
+	public void add_parameter (string key, string value) {
+		parameters.insert (
+			GLib.Uri.escape_string (key, "[]"),
+			GLib.Uri.escape_string (value)
+		);
+	}
+
+	public bool remove_parameter (string key) {
+		if (!this.parameters.contains (key)) return false;
```
bugaevc: But `parameters` stores pre-escaped keys?

src/Services/Network/RequestV2.vala
```diff
@@ -0,0 +1,158 @@
...
+			return p_key == key || p_key == @"$key[]";
+		}) > 0;
+	}
+
+	public void add_parameter_array (string key, string[] values) {
+		if (values.length == 0) return;
+
+		string final_key = key;
+		if (!final_key.has_suffix ("[]")) final_key = @"$key[]";
+		foreach (string value in values) {
```
bugaevc:
```suggestion
		foreach (unowned string value in values) {
```

src/Services/Network/RequestV2.vala
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	public void set_body_from_json (Json.Builder json_builder) {
+		Json.Generator generator = new Json.Generator ();
+		generator.set_root (json_builder.get_root ());
+		set_body ("application/json", new Bytes.take (generator.to_data (null).data));
+	}
+
+	public async bool exec (out GLib.InputStream in_stream, out Soup.MessageHeaders response_headers) throws GLib.Error, Oopsie {
+		if (this.cancellable != null && !this.cancellable.is_cancelled ()) this.cancellable.cancel ();
```
bugaevc:
```suggestion
		this.cancellable?.cancel ();
```

`GLib.Cancellable.cancel ()` already performs the already-cancelled check.

src/Views/Timeline.vala
```diff
@@ -212,33 +212,46 @@ public class Tuba.Views.Timeline : AccountHolder, Streamable, Views.ContentBase
...
+		};
+
+		if (page_next == null)
+			req.add_parameter ("limit", settings.timeline_page_size.clamp (this.batch_size_min, 40).to_string ());
+
+		request_async.begin (req, (obj, res) => {
+			try {
+				request_async.end (res);
+			} catch (GLib.Error e) {
+				on_error (e.code, e.message);
```
bugaevc: Perhaps move this to a `catch` block in `request_async` itself?

I have a preference for using the `async`/`yield` syntax as much as possible, and only spawning async tasks through `.begin ()` without a completion callback. In other words, if you have a completion, it should have been control flow in that asynchronous method.

bugaevc (commented):
Some quick notes

src/Services/Network/Network.vala
GeopJr: Not intentional!

GeopJr:
Thanks for the review!

Maybe the API should change again and return the inputstream instead of outing it, now that we don't need to return bool

src/Services/Network/RequestV2.vala
GeopJr: Removed the `is_cancelled` check, but fwiw, I'm hesitant on using optional chaining in vala due to #1002 and f5bd1c93f50eab09b09bca4032115fe893a0199f (couldn't reliably reproduce to report upstream)

bugaevc:
Yeah that makes sense too.

src/Views/Timeline.vala
```diff
@@ -224,6 +224,11 @@ public class Tuba.Views.Timeline : AccountHolder, Streamable, Views.ContentBase
 			try {
 				request_async.end (res);
 			} catch (GLib.Error e) {
+				if (e is GLib.IOError.CANCELLED) {
+					debug ("Message is cancelled.");
```
bugaevc: Vala has syntax for this 😀

```vala
try {
} catch (GLib.IOError.CANCELLED e) {
    ...
} catch (GLib.Error e) {
    ...
}
```

GeopJr:
Moving forward with replacing all instances, something that caught my attention is vala's construction order:

```vala
// Home.vala, moving the refreshing away from idle and to just the async function
	construct {
		url = "/api/v1/timelines/home";
		label = _("Home");
		icon = "tuba-user-home-symbolic";
```

fails with url == null

but

```vala
    public Home () {
        Object (
            url: "/api/v1/timelines/home",
	    label: _("Home")
        );
    }
```

works. At the same, the Notifications.vala url depends on the account, which is not available during `Notifications () {`, but leaving it as is doesn't cause any url == null

:thinking: Personally, I'd expect construct to run before using the Object at all. I know I'm wrong now, but I thought it was like this: construct -> Home () -> Object ()

GeopJr:
More annoyances:

```vala
public Views.ContentBase add_timeline_tab (string label, string icon, string url, Type accepts, string? empty_state_title = null, string? empty_state_icon = null) {
		var tab = new Views.Accounts () {
			url = url,
			label = label,
			icon = icon,
			accepts = accepts
		};
```
url == null, but introducing this works:

```vala
public class Tuba.Views.Accounts : Views.Timeline {
	public Accounts.prefilled (string url, string label, string icon) {
		Object (
			url: url,
			label: label,
			icon: icon
		);
	}
```

bugaevc:
On top of GObject's already complicated construction/initialization flow, Vala adds its own layer of confusion :sob: I hate everything about the way it's done, and it's unlikely that I could change much here about Vala semantics (especially given [the state of Vala maintenance](https://discourse.gnome.org/t/lets-move-vala-forward/34595), which makes it hard to get any non-trivial change merged into Vala).

You would think `new Type () { prop1= value1, prop2 = value2 };` syntax calls `g_object_new (THE_TYPE, "prop1", value1, "prop2", value2, NULL)`, right? But it doesn't, instead it calls the regular constructor (`the_type_new ()`) followed by setting the individual properties.

Properties whose values are visible at `constsruct { }` time are the ones that both:
* are "construct properties", i.e. marked `construct` or `construct set`
* were passed to the `g_object_new` call, i.e. `Object ()` chain-up in Vala

What you mean about `Home.vala` and URL being null, I couldn't really follow (more details please?), but possibly you're reading the value of the `url` property (or starting an async operation which does just that) from the `construct` block of a parent class, but setting that property from the `construct` block of a derived class? The derived class's `construct` runs later than the parent's, hence the value is not yet set at that point — if that's what you're doing & running into.

From the GObject standpoint, the right thing to do is:
1. Passing as many properties as possible into the `g_object_new` call, as opposed to creating an instance first and setting properties later — i.e. the opposite of what `new Type () { prop = value }` syntax does in Vala. You could achieve this in Vala by calling `g_object_new` more explicitly (`Object.@new (typeof (Type), prop: value)`), but this is somewhat ugly; alternatively, creating special-case constructors with `Object ()` chain-ups, as in your `Accounts.prefilled` example — also cumbersome.
2. Generally, doing/starting various things _in property setters_, when a value is set for a property, and not "upon creation". E.g. I imagine a request could be started from the `url` property's setter. This way, if the object does get created without values for all these properties, those "various things" do happen when the values do get set. It's also a good opportunity to make the property re-settable, i.e. start the setter by undoing whatever setup it did for the old value (e.g. cancelling an old request, if any, when the URL changes). For example, see how `Gtk.DirectoryList` reacts when its `file` property is set (`gtk_directory_list_set_file`).
3. When doing things in setters is not feasible, marking properties `construct` and doing things in the `constructed` vfunc; Vala's `construct` blocks actually translate to the `constructor` vfunc, which is worse, but close enough.

Here's a piece of code for you to play with:

```vala
#! /usr/bin/vala

public class Base : Object {
	public int some_prop {
		set { print ("Setting some_prop to %d\n", value); }
	}
	public int some_construct_prop {
		construct set { print ("Setting some_construct_prop to %d\n", value); }
	}

	public Base () {
		print ("Base: before chain-up\n");
		Object (some_prop: 42);
		print ("Base: after chain-up\n");
	}

	construct {
		print ("Base: construct block\n");
	}
}

public class Derived : Base {
	public Derived () {
		print ("Derived: before chain-up\n");
		base ();
		print ("Derived: after chain-up\n");
	}

	construct {
		print ("Derived: constrct block\n");
	}
}

void main () {
	new Derived ();
}
```

I get:

```
Derived: before chain-up
Base: before chain-up
Setting some_construct_prop to 0
Base: construct block
Derived: constrct block
Setting some_prop to 42
Base: after chain-up
Derived: after chain-up
```

GeopJr:
> What you mean about Home.vala and URL being null, I couldn't really follow (more details please?), but possibly you're reading the value of the url property (or starting an async operation which does just that) from the construct block of a parent class, but setting that property from the construct block of a derived class? The derived class's construct runs later than the parent's, hence the value is not yet set at that point — if that's what you're doing & running into.

Timeline.vala inherits from AccountHolder which when constructed refreshes the timeline (initial fetch), and that's where url is null. Prior to changing everything to the async one, it would do the request in idle_add (also why we don't come across it in the current state of this PR) and :crossed_fingers: we were lucky it finished after the constructions. But now it seems that running the async request directly is faster. (FWIW, url is `construct set`)

> I imagine a request could be started from the url property's setter

That sounds good, but currently it's more tied to the account (so changing accounts refreshes the timelines as the new account) and there's also other things in some timelines (from pagination to filters and api versions; that might not need immediate fetching). I'll see if I can do it without breaking too many things.

> When doing things in setters is not feasible, marking properties construct and doing things in the constructed vfunc; Vala's construct blocks actually translate to the constructor vfunc, which is worse, but close enough.

I think I'll go with this. Right now, I've been literally just moving them to:
```vala
    public Home () {
        Object (
            url: "/api/v1/timelines/home",
            label: _("Home")
        );
    }
```

`constructed` looks less random "let's hope the order is right with this one".

> Here's a piece of code for you to play with:

as suspected, I really did think it was `base construct -> derived construct -> derived before chain-up -> base before chain-up -> base after chain-up -> derived after chain-up` :p I wonder what other things are wrong in Tuba from that alone...

---

> especially given [the state of Vala maintenance](https://discourse.gnome.org/t/lets-move-vala-forward/34595), which makes it hard to get any non-trivial change merged into Vala

I read that, and honestly, what happens next depends on Rico who hasn't really acknowledged it as far as I can tell. Outside of GNOME/Elementary, maybe the Frida / NowSecure people would be interested in moving Vala forward? (since it's written in Vala)

Speaking of Elementary, they use lambdas a lot... https://github.com/search?q=org%3Aelementary%20%3D%3E&type=code one can only imagine the memory benefits they could get by getting those ref loops fixed

